PR #15, opened 2024-08-20
Ethan/cd

AgainstEntropy:


AgainstEntropy:
> could you make sure all layers=-1 is changed, nexa_service.py is not changed yet.

fixed

PR #28, opened 2024-08-22
unified cpu and gpu build

xyyimian:


AgainstEntropy:
We need to put stable-diffusion before llama-cpp to make CUDA works for stable-diffusion

PR #37, opened 2024-08-22
Update README and add -j option

zhiyuan8:


zhiyuan8:
windows CI has some path issue @xyyimian

PR #50, opened 2024-08-27
add a demo for `financial-advisor`

JoyboyBrian:


examples/financial-advisor/app.py
```diff
@@ -0,0 +1,278 @@
+import sys
+import os
+import streamlit as st
+from typing import Iterator
+import subprocess
+import json
+import shutil
+import pdfplumber
+from sentence_transformers import SentenceTransformer
```
zhiyuan8, line 6: can we use Nexa's own embedding inteface? Do not use SentenceTransformers. You can check with @xyyimian

examples/financial-advisor/utils/text_generator.py
```diff
@@ -0,0 +1,130 @@
+import os
+import faiss
+import numpy as np
+import logging
+from nexa.gguf import NexaTextInference
+from langchain_community.embeddings import HuggingFaceEmbeddings
```
zhiyuan8: Do not use 3rd party embedding

zhiyuan8 (changes requested):

PR #80, opened 2024-09-10
update llama.cpp and sd.cpp to latest

zhiyuan8:


zhiyuan8:
stable-diffucion.cpp macOS fails, investigating

PR #90, opened 2024-09-14
support ROCm wheel build

zhycheng614:


.github/workflows/build-wheels-rocm-linux.yaml
```diff
@@ -0,0 +1,87 @@
...
+      run:
+        shell: pwsh
+
+    steps:
+      - name: Define Job Output
+        id: set-matrix
+        run: |
+          $matrix = @{
+              'os' = @('ubuntu-22.04')
+              'pyver' = @("3.10", "3.11", "3.12")
```
zhiyuan8: See https://github.com/NexaAI/nexa-sdk/blob/main/.github/workflows/build-wheels-cuda-win.yaml#L23
cover more python version

.github/workflows/build-wheels-rocm-linux.yaml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Build Wheel
+        run: |
+          export ROCM_PATH=/opt/rocm
+          export HIP_PATH=$ROCM_PATH/hip
+          export PATH=$ROCM_PATH/bin:$PATH
+          export LD_LIBRARY_PATH=$ROCM_PATH/lib:$LD_LIBRARY_PATH
+
+          export VERBOSE=1
+          export CMAKE_BUILD_PARALLEL_LEVEL=$(nproc)
+          export CMAKE_ARGS='-DGGML_HIPBLAS=ON -DCMAKE_HIP_COMPILER="$(hipconfig -l)/clang"'
```
zhiyuan8, line 72: Have you verified stable-diffusion.cpp?
check
https://github.com/leejet/stable-diffusion.cpp?tab=readme-ov-file#using-hipblas

zhiyuan8 (changes requested):
Also need to update README.md for RoCM
And check stable-diffusion.cpp RoCM flags and combine with llama.cpp

zhiyuan8 (approved):

Issue #136, opened 2024-10-02
[FEATURE] <title> run_type "Multimodal" missing in nexa_service.py


tigerrabbit:
### Issue Description

when running server llava1.6-mistral - unable to load model
nexa server llava1.6-vicuna
objc[12203]: Class GGMLMetalClass is implemented in both /Applications/Nexa.app/Contents/Frameworks/libggml_llama.dylib (0x12083c260) and /Applications/Nexa.app/Contents/Frameworks/nexa/gguf/lib/libstable-diffusion.dylib (0x1230a0260). One of the two will be used. Which one is undefined.
INFO:     Started server process [12203]
INFO:     Waiting for application startup.
2024-10-02 12:55:30,162 - INFO - Model Path: llava1.6-vicuna
Model llava-v1.6-vicuna-7b:model-q4_0 already exists at /Users/gg/.cache/nexa/hub/official/llava-v1.6-vicuna-7b/model-q4_0.gguf
ERROR:    Traceback (most recent call last):
  File "starlette/routing.py", line 693, in lifespan
  File "starlette/routing.py", line 569, in __aenter__
  File "starlette/routing.py", line 670, in startup
  File "nexa/gguf/server/nexa_service.py", line 343, in startup_event
  File "nexa/gguf/server/nexa_service.py", line 203, in load_model
ValueError: Model llava1.6-vicuna not found in Model Hub

ERROR:    Application startup failed. Exiting.

### Steps to Reproduce

nexaai server llava1.6-mistral and compare to nexaai server llava1.6-mistral

### OS

mac os latest

### Python Version

3.12.6

### Nexa SDK Version

latest

### GPU (if using one)

mac metal

Davidqian123:
For now, we don't support multimodal models server, you can try `nexa run llava1.6-mistral`.
If you want to host local server for multimodal models, we will improve our sdk to support multimodal models server in the future.

JGalego:
+1 for multimodal models server support

tigerrabbit:
I am in love with the product.  I was impressed with the coding when I dug into this using the editable install option.  I am going to be keeping up with this and look forward to this option.  I am sure there are some gotchas that I am not even aware of in implementing this feature.

Davidqian123:
we will include this in our roadmap

zhiyuan8:
> +1 for multimodal models server support

Thank you, we are priotizing this and it expects to be delivered in next 2 weeks.

[...]

PR #154, opened 2024-10-08
Feature Implementation for issue #136 : supports vlm requests for /chat/completions api

zhycheng614:
- updates on the existing api: /chat/completions api to be compatible with OpenAI api
- supports streaming
- running model from local path supported, with --local_path and --model_type flags provided

### DIFF
--- a/SERVER.md
+++ b/SERVER.md
@@ -22,14 +22,15 @@ nexa server gemma
 nexa server llama2-function-calling
 nexa server sd1-5
 nexa server faster-whipser-large
+nexa server ../models/llava-v1.6-vicuna-7b/ -lp -mt MULTIMODAL
 ```
 
 By default, `nexa server` will run gguf models. To run onnx models, simply add `onnx` after `nexa server`.
 
 ## API Endpoints
 
-
 ### 1. Text Generation: <code>/v1/completions</code>
+
 Generates text based on a single prompt.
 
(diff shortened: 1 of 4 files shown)


SERVER.md
```diff
@@ -54,13 +55,46 @@ Generates text based on a single prompt.
...
+
 Handles chat completions with support for conversation history.
 
 #### Request body:
 
+Multimodal models (VLM):
+
+```json
+{
+  "model": "anything",
```
zhiyuan8, line 70: Please use an actual model name, developers may just want to copy it and it should work

nexa/cli/entry.py
```diff
@@ -15,13 +15,41 @@ def run_ggml_inference(args):
     if model_type:
         run_type = ModelType[model_type].value
 
+    def choose_file(local_path, file_type):
+        """ Helper function for Multimodal inference only: select the model and projector ggufs from the local_path. """
```
zhiyuan8, line 19: why don't we split proj and model file into two arguments? We might want to add a flag "--proj" or "-p" for projector. What if I mistakenly write proj at 1st position?

zhiyuan8: please confirm this UX with product @alanzhuly

zhiyuan8 (approved):

PR #171, opened 2024-10-18
eval without server

JoyboyBrian:


pyproject.toml
```diff
@@ -38,6 +38,7 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
```
zhiyuan8: you need to add python 3.13 for github actions
I have added it, please rebase

zhiyuan8 (changes requested):

pyproject.toml
JoyboyBrian: Done

zhiyuan8 (approved):

PR #203, opened 2024-11-01
[New Feature] Swift Package & Examples

JoyboyBrian:


Package.swift
```diff
@@ -0,0 +1,32 @@
+// swift-tools-version: 6.0
+
+import PackageDescription
+
+let package = Package(
+    name: "NexaSwift",
+    platforms: [
+        .macOS(.v15),
+        .iOS(.v18),
```
zhiyuan8, line 9: is it possible to lower the iOS and macOS version requirements?
llama.cpp writes iOS as v14
https://github.com/ggerganov/llama.cpp/blob/master/Package.swift#L51-L53

zhiyuan8 (changes requested):

examples/swift-test/Shared/ViewModel.swift
```diff
@@ -0,0 +1,57 @@
...
+class ViewModel {
+    let nexaSwift: NexaTextInference
+    var result = ""
+    var usingStream = true
+    private var messages: [ChatCompletionRequestMessage] = []
+    private let maxHistory = 1
+    private var cancallable: Set<AnyCancellable> = []
+    
+    init() {
+        let configuration = Configuration(maxNewToken: 128, stopTokens: ["<nexa_end>"])
```
zhiyuan8: Add notes why putting early stopping here

swift/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+```swift
+for try await response in await nexaSwift.createCompletionStream(for: prompt) {
+    print(response.choices[0].text)
+}
+```
+
+## Quick Start
+
+Open the [swift test project](../examples/swift-test/) folder in Xcode and run the project.
```
zhiyuan8, line 80: Add a bit explanation about using Octopus-V2 for iOS

swift/Sources/NexaSwift/General.swift
```diff
@@ -0,0 +1,123 @@
...
+    "Llama2-7b-function-calling:q3_K_M": "Llama2-7b-function-calling:q3_K_M",
+    "Llama2-7b-function-calling:q3_K_S": "Llama2-7b-function-calling:q3_K_S",
+    "Llama2-7b-function-calling:q4_K_M": "Llama2-7b-function-calling:q4_K_M",
+    "Llama2-7b-function-calling:q4_K_S": "Llama2-7b-function-calling:q4_K_S",
+    "Llama2-7b-function-calling:q5_K_M": "Llama2-7b-function-calling:q5_K_M",
+    "Llama2-7b-function-calling:q5_K_S": "Llama2-7b-function-calling:q5_K_S",
+    "Llama2-7b-function-calling:q6_K": "Llama2-7b-function-calling:q6_K",
+    "Llama2-7b-function-calling:q8_0": "Llama2-7b-function-calling:q8_0",
+]
+let NEXA_RUN_MODEL_MAP_FLUX:[String: String] = [
```
zhiyuan8: image generation models cannot work, remove

swift/Sources/NexaSwift/General.swift
```diff
@@ -0,0 +1,123 @@
...
+    "sd2-1": "stable-diffusion-v2-1:q4_0",
+    "sd3": "stable-diffusion-3-medium:q4_0",
+    "sdxl-turbo": "sdxl-turbo:q8_0",
+    "flux": "FLUX.1-schnell:q4_0",
+    "lcm-dreamshaper": "lcm-dreamshaper-v7:fp16",
+    "anything-lcm": "anything-v30-LCM:fp16",
+    "hassaku-lcm": "hassaku-hentai-model-v13-LCM:fp16",
+]
+
+let NEXA_RUN_MODEL_MAP_VOICE:[String: String] = [
```
zhiyuan8: ASR models cannot work, remove

swift/Sources/NexaSwift/General.swift
```diff
@@ -0,0 +1,123 @@
...
+    "dolphin-mistral": "dolphin-2.8-mistral-7b:q4_0",
+    "phi2": "Phi-2:q4_0",
+    "phi3": "Phi-3-mini-128k-instruct:q4_0",
+    "phi3.5": "Phi-3.5-mini-instruct:q4_0",
+    "llama2-uncensored": "Llama2-7b-chat-uncensored:q4_0",
+    "llama3-uncensored": "Llama3-8B-Lexi-Uncensored:q4_K_M",
+    "openelm": "OpenELM-3B:q4_K_M"
+]
+
+let NEXA_RUN_MODEL_MAP_VLM: [String: String] = [
```
zhiyuan8: Only keep model mappings currently work on iOS

Package.swift
JoyboyBrian: Our package uses some functions (like `init(validating:as:)`) that are only available in newer OS versions, which sets our current minimum requirements. We can consider backward compatibility in future versions based on community needs.

zhiyuan8 (changes requested):

examples/swift-test/Shared/ViewModel.swift
JoyboyBrian: This is a demostration how we can use NexaSwift package in Swift. I will change the default model to a more general one, i.e. gemma2 or llama3.2. I will also add another example project show how to use octopus-v2 model and use the functional token to trigger actions later.

swift/README.md
JoyboyBrian: Will add the explanation in the example project folder regarding octopus-v2.

swift/Sources/NexaSwift/General.swift
JoyboyBrian: We are not using the mapping for now, can be deleted.

swift/Sources/NexaSwift/General.swift
JoyboyBrian: We are not using the mapping for now, can be deleted.

swift/Sources/NexaSwift/General.swift
JoyboyBrian: We are not using the mapping for now, can be deleted.

zhiyuan8 (approved):

PR #205, opened 2024-11-01
fix load llava library error

Davidqian123:


nexa/gguf/llama/llava_cpp.py
```diff
@@ -33,7 +33,7 @@
 from nexa.gguf.lib_utils import load_library
 
 # Specify the base name of the shared library to load
-_libllava_base_name = "llava"
```
vinovo, line 36: Did we check that other OS all have _shared suffix? I though the problem was only on windows?

Davidqian123: I have checked that all os have _shared suffix

zhiyuan8 (approved):

PR #209, opened 2024-11-03
Enable AudioLM works for Qwen2Audio & nano-omni

zhiyuan8:


zhiyuan8:
img_v3_02ga_74266a26-ca66-4ef4-9fcb-9c6eb39e6c1h
works for AMD laptop

zhiyuan8:
CI works after removing C++20 features from omni-audio implementations

PR #253, opened 2024-11-19
support modelscope hub

mi804:
Add modelscope hub support.


zhiyuan8:
Thanks for adding this option

PR #286, opened 2024-11-25
add dependency for modelscope

mi804:
We find that the latested prebuilt whl or Executable Installer do not contain the modelscope module.
To import modelscope, I add the `modelscope` dependency in the pyproject.toml and I guess this may install the modelscope module when building `nexaai`. If there are any other places where the modelscope dependencies need to be added, please let me know. And thank you for updating the whl and Executable Installer to the version that support model downloading from ModelScope.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,6 +22,7 @@ dependencies = [
     "pydantic",
     "pillow",
     "huggingface_hub",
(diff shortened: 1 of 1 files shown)


zhiyuan8:
@mi804 Thanks for your contribution

PR #384, opened 2025-02-10
feat: optimized logits logic in nexa eval, 50 times faster when nexa …

zhycheng614:
…eval now


zhiyuan8 (approved):
@zhycheng614  please update this change in readme, emphasize we have supported 50X faster GGUF benchmark, and provide a sample code, versus original lm_eval

Issue #442, opened 2025-05-08
[QUESTION] <title>  no member named 'sparams' in


oznfc:
### Question or Issue

when ı trid run example app (app-java)  getting this error : 
```
workdir/tauri/nexa-works/test/nexa-sdk/android/llama/src/main/cpp/llava-android.cpp:154:13: error: no member named 'sparams' in 'common_params'
  154 |     params->sparams.top_p = top_p;
      |     ~~~~~~  ^
```

### OS

ubuntu 24

maxma1973-github:
Hi oznfc,

Have you addressed the issue, I faced same issue when I build the android app.

Thanks and best regards,
Max

Siddhesh2377:
Found the Fix GUYSS............>!!!!

explained

```
struct common_params {
    ...
    struct common_params_sampling    sampling;
    struct common_params_speculative speculative;
    struct common_params_vocoder     vocoder;
    ...
};
```


So use 

```
params->sampling.top_k
params->sampling.top_p
params->sampling.temp

```

keithem:
Thank u

Keith cox

MaokunZhang:
👋 Important Update: Nexa-SDK was fully rewritten on July 23, 2025, paving the way for exciting new features. This issue relates to a legacy version (v1) and has now been archived with the label legacy.

👉 Still experiencing this issue in the new version? Please open a new issue
🔗 Explore what's new in Nexa-SDK v2: [README](https://github.com/NexaAI/nexa-sdk)

Thanks for helping us make Nexa-SDK even better! 🚀✨

Issue #444, opened 2025-06-17
The common_params struct does not define a member named sparams. Instead, it defines a valid nested struct named sampling of type common_params_sampling


shadhin-music:
In the original codebase, the **common_params** struct was incorrectly accessed using **params->sparams**, as seen in files like llava-android.cpp. This led to compile-time errors like:

**error: no member named '**sparams**' in '**common_params**'
params->sparams.top_k = top_k;
        ^~~~~~~**

The **common_params** struct does not define a member named **sparams**. Instead, it defines a valid nested struct named **sampling** of type **common_params_sampling**:

struct common_params {
    ...
    struct common_params_sampling sampling;
    ...
};

✅ Fix Applied

All usages of params->sparams across .cpp source files have been replaced with params->sampling to reflect the correct member.
✅ Example Before Fix (Incorrect)

params->sparams.top_k = top_k;
params->sparams.top_p = top_p;
params->sparams.temp  = temp;

✅ Example After Fix (Correct)

params->sampling.top_k = top_k;
params->sampling.top_p = top_p;
params->sampling.temp  = temp;

MaokunZhang:
👋 Important Update: Nexa-SDK was fully rewritten on July 23, 2025, paving the way for exciting new features. This issue relates to a legacy version (v1) and has now been archived with the label legacy.

👉 Still experiencing this issue in the new version? Please open a new issue
🔗 Explore what's new in Nexa-SDK v2: [README](https://github.com/NexaAI/nexa-sdk)

Thanks for helping us make Nexa-SDK even better! 🚀✨

PR #445, opened 2025-06-17
Fixed all incorrect references to params->sparams by replacing them w…

rjfahad44:
Fixed all incorrect references to params->sparams by replacing them with params->sampling to match the actual struct member defined in common_params.


iwr-redmond:
cc @Davidqian123 

Closes: #442, #444

PR #487, opened 2025-08-26
fix: add image_gen to infer helper prompt

vinovo:


mengshengwu (approved):

mengshengwu (commented):
`output` also need change

PR #507, opened 2025-09-02
feat: build windows launcher using golang

lxsong1:


runner/release/windows/launcher.go
```diff
@@ -0,0 +1,14 @@
+//go:build windows
+
+package main
+
+import (
+	"os/exec"
+	"syscall"
+)
+
+func main() {
```
RemiliaForever, line 10: can this merge to nexa-launcher？

PR #545, opened 2025-09-17
feat: swift-binding-release

YangXianda007:


Package.resolved
RemiliaForever, line 1: should this put on root?

Package.swift
RemiliaForever, line 1: same as above

.gitignore
```diff
@@ -70,4 +70,10 @@ tmp/
 __debug_bin
 
 # TLS certificates
-*.pem
\ No newline at end of file
+*.pem
```
mengshengwu, line 73: It would be better to create a new .gitignore for ios instead of putting it in the root directory

PR #555, opened 2025-09-19
fix: allow .nexa and .gguf to be in the same repo

vinovo:


RemiliaForever:
The `chooseFiles` function is getting more and more complex. I suggest solving the self-hosted repo by extending the fields in repo’s `nexa.manifest`.

RemiliaForever (approved):

PR #569, opened 2025-09-22
feat(cli): add image max length parameter to generation config

mengshengwu:
- Introduced `imageMaxLength` to the `infer` command for better control over image input size.
- Updated `GenerationConfig` struct to include `ImageMaxLength`.
- Adjusted related function calls to accommodate the new parameter.

### DIFF
--- a/runner/cmd/nexa-cli/infer.go
+++ b/runner/cmd/nexa-cli/infer.go
@@ -30,22 +30,23 @@ const modelLoadFailMsg = `⚠️ Oops. Model failed to load.
 
 var (
 	// disableStream *bool // reuse in run.go
-	ngl          int32
-	maxTokens    int32
-	enableThink  bool
-	hideThink    bool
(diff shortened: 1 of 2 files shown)


runner/cmd/nexa-cli/infer.go
```diff
@@ -64,6 +65,7 @@ func infer() *cobra.Command {
 	inferCmd.Flags().SortFlags = false
 	inferCmd.Flags().Int32VarP(&ngl, "ngl", "n", 999, "[llm|vlm] num of layers pass to gpu")
 	inferCmd.Flags().Int32VarP(&maxTokens, "max-tokens", "", 2048, "[llm|vlm] max tokens")
+	inferCmd.Flags().Int32VarP(&imageMaxLength, "max-image-length", "", 512, "[vlm] max image length")
```
RemiliaForever: make variable name same as param

PR #588, opened 2025-09-25
feat(runner): add symlink creation for model directory for sdxl-turbo

mengshengwu:


RemiliaForever:
please move the model-specific logic to nexasdk-bridge.

PR #593, opened 2025-09-26
feat: add image generation support for server

mengshengwu:


runner/server/handler/image.go
```diff
@@ -0,0 +1,332 @@
...
+
+	"github.com/NexaAI/nexa-sdk/runner/internal/store"
+	"github.com/NexaAI/nexa-sdk/runner/internal/types"
+	nexa_sdk "github.com/NexaAI/nexa-sdk/runner/nexa-sdk"
+	"github.com/NexaAI/nexa-sdk/runner/server/service"
+)
+
+// ImageGenerationRequest represents the request body for image generation
+// Following OpenAI DALL-E 3 API specification
+type ImageGenerationRequest struct {
```
RemiliaForever: use struct in `openai` lib instead

runner/server/handler/image.go
```diff
@@ -0,0 +1,332 @@
...
+	setDefaultImageGenerationParams(&param)
+
+	// Validate parameters
+	if err := validateImageGenerationParams(param); err != nil {
+		slog.Error("Invalid image generation parameters", "error", err, "params", param)
+		c.JSON(http.StatusBadRequest, map[string]any{"error": err.Error()})
+		return
+	}
+
+	// Get model manifest
```
RemiliaForever: move model type check to keepalive is better

runner/server/handler/image.go
```diff
@@ -0,0 +1,332 @@
...
+	imageData, err := io.ReadAll(file)
+	if err != nil {
+		return "", fmt.Errorf("failed to read image file: %v", err)
+	}
+
+	// Encode to base64
+	base64String := base64.StdEncoding.EncodeToString(imageData)
+
+	// Determine MIME type based on file extension
+	var mimeType string
```
RemiliaForever: maybe detect content instead of detect file extension

RemiliaForever (changes requested):

RemiliaForever (approved):

PR #605, opened 2025-09-28
feat(handler): add base64 image handling for chat completions

mengshengwu:


runner/server/handler/completion.go
```diff
@@ -319,11 +320,26 @@ func chatCompletionsVLM(c *gin.Context, param ChatCompletionRequest) {
 						Text: *ct.GetText(),
 					})
 				case "image_url":
+					imageURL := ct.GetImageURL().URL
+					if isBase64ImageData(imageURL) {
```
RemiliaForever: add a function like `parseURL`, support `http[s]://`,`file:`,`data:`

PR #631, opened 2025-10-03
fix: set up nexa keep cache in nexa run

vinovo:


RemiliaForever (approved):

RemiliaForever:
`run` warm up missing some fields, and `serve` missing system prompt detect, i will fix later.

PR #654, opened 2025-10-09
feat(infer): add --no-interactive flag for non-interactive mode support

mengshengwu:


runner/cmd/nexa-cli/infer.go
```diff
@@ -349,6 +352,45 @@ func inferLLM(manifest *types.ModelManifest, quant string) {
 		// return
 	}
 
+	if noInteractive {
```
RemiliaForever, line 357: please do not copy code, which make code hard to maintain

runner/cmd/nexa-cli/infer.go
```diff
@@ -84,6 +85,7 @@ var (
 	llmFlags = func() *pflag.FlagSet {
 		llmFlags := pflag.NewFlagSet("LLM/VLM Model", pflag.ExitOnError)
 		llmFlags.SortFlags = false
+		llmFlags.BoolVarP(&noInteractive, "no-interactive", "", false, "disable interactive mode")
```
RemiliaForever, line 88: missing prompt param

RemiliaForever, line 88: does prompt conflict with input?

mengshengwu: As we design before:
input pass file
prompt pass text directly

RemiliaForever: that's confused, prompt only work on no-interactive mode.

and both input and prompt do not support image/audio

RemiliaForever: multiple prompts work as multiround inputs would be better

PR #699, opened 2025-10-17
refactor:  use model id for vlm example

mengshengwu:
Because upstream does not support passing modelfile directly at the moment


mengshengwu:
duplicated #708

PR #707, opened 2025-10-19
Revert "chore: bump up nexa bridge version to v1.0.21-rc4"

keithem:
Reverts NexaAI/nexa-sdk#706


RemiliaForever:
Is there a specific reason for reverting this?

PR #719, opened 2025-10-21
nexa-cli: print success message after download completes in nexa pull

ximing-crypto:


runner/cmd/nexa-cli/model.go
```diff
@@ -242,6 +242,7 @@ func pullModel(name string) error {
 			bar.Clear()
 			fmt.Println(render.GetTheme().Error.Sprintf("Error: %s", err))
 		}
+		fmt.Println(render.GetTheme().Success.Sprintf("Download success！"))
```
mengshengwu: No output means success — this tips seems redundant.

runner/cmd/nexa-cli/model.go
```diff
@@ -287,6 +288,7 @@ func pullModel(name string) error {
...
```
RemiliaForever: Instead of adding print statements in both if branches, it's better to add a single print statement after the if branch.

RemiliaForever: Consider adding emojis to the print output, keeping the format consistent with `remove` and `clean`.

runner/cmd/nexa-cli/model.go
ximing-crypto: done

runner/cmd/nexa-cli/model.go
ximing-crypto: done

ximing-crypto: done

RemiliaForever (approved):

PR #769, opened 2025-10-27
feat: add notebook for mac and win(x64)

mengshengwu:


zhiyuan8:
@mengshengwu mac models, such as MLX & GGUF models, do not need token

PR #801, opened 2025-10-30
fix(completion): change `choices` from null to [] for compatibility with some OpenAI-compatible API clients, such as Cherry Studio

wenyifancc:
Some OpenAI-compatible API clients will report an error when the choices field in the response content is null, such as Cherry Studio.


mengshengwu (approved):

RemiliaForever:
Thanks!

PR #834, opened 2025-11-08
fix: Replace TryLock with blocking Lock in GIL middleware to prevent 429 errors

### LINKED ISSUE #833: CRITICAL: NexaAI SDK v0.2.57 - Excessive 429 "Too Many Requests" Throttling for Local Embedding Requests
## 🚨 Bug Description
**NexaAI SDK Version**: v0.2.57 (Bridge: v1.0.28-rc5)
**Environment**: Apple Silicon Mac, macOS Sonoma, localhost:8010

### Issue Summary
The NexaAI server is experiencing excessive HTTP 429 "Too Many Requests" errors when processing local embedding requests, even for single concurrent requests. This causes severe performance degradation and makes the embedding service unreliable for production use.

### Expected Behavior
Local single embedding requests should succeed immediately with 200 status, without any rate limiting applied to localhost connections.

### Actual Behavior
- Hundreds of 429 errors appear in server logs
- Failed requests return immediately with 429 status
- Successful requests show 20-750ms response times
- Overall throughput severely degraded by SDK-level throttling

### Steps to Reproduce
1. Start NexaAI server: `nexa serve --host 0.0.0.0:8010`
2. Make single embedding request: 
   ```bash
   curl -X POST http://localhost:8010/v1/embeddings \
       -H "Content-Type: application/json" \
       -d '{"model":"embeddinggemma-300m-npu","input":"test"}'
   ``` {data-source-line="51"}
3. Observe immediate 429 response in logs: `tail -f /tmp/nexa-server.log | grep "429"`

### Log Evidence
[GIN] 2025/11/07 - 22:17:01 | 500 | 220.125µs | 127.0.0.1 | POST "/v1/embeddings"
[GIN] 2025/11/07 - 22:17:01 | 429 | 201.25µs | 127.0.0.1 | POST "/v1/embeddings"
[GIN] 2025/11/07 - 22:17:01 | 500 | 286µs | 127.0.0.1 | POST "/v1/embeddings"

artik0din:
## 🐛 Bug Fix

Fixes #833

## Problem

The GIL middleware was causing excessive HTTP 429 "Too Many Requests" errors for local embedding requests, even with single concurrent requests. The root cause was `TryLock()` returning immediately when the lock is held, causing legitimate requests to fail instead of waiting.

### Observed Issues:
- Hundreds of 429 errors in server logs
- Failed requests with 200µs response times (too fast = not processed)
- Severe throughput degradation
- Unreliable embedding service for production use

## Solution

Replaced `sync.Mutex.TryLock()` with `sync.Mutex.Lock()` in the GIL middleware:

- **Before**: Requests fail immediately with 429 if lock is held
- **After**: Requests queue and wait for lock availability

This maintains the same serialization guarantees while eliminating false-positive rate limiting.

## Changes

- Modified `runner/server/middleware/gil.go`
- Replaced non-blocking `TryLock()` with blocking `Lock()`
- Simplified error handling (no 429 check needed)

## Testing

**Environment:**
- macOS Sonoma (Apple Silicon)
- NexaAI SDK v0.2.57
- localhost:8010

**Before fix:**
curl -X POST localhost:8010/v1/embeddings \
  -H "Content-Type: application/json" \
  -d '{"model":"embeddinggemma-300m-npu","input":"test"}'
# Result: HTTP 429 errors frequently**After fix:**
# Multiple concurrent requests
for i in {1..5}; do
  curl -X POST localhost:8010/v1/embeddings \
    -H "Content-Type: application/json" \
    -d '{"model":"embeddinggemma-300m-npu","input":"test '$i'"}' &
done
# Result: All requests succeed with HTTP 200## Impact

- ✅ **Reliability**: Eliminates false-positive 429 errors
- ✅ **Performance**: Minimal latency increase (requests wait instead of fail)
- ✅ **Compatibility**: No breaking changes
- ✅ **Serialization**: Maintains existing concurrency control

## Related

- Issue: #833
- Component: Server middleware (GIL)

### DIFF
--- a/runner/server/middleware/gil.go
+++ b/runner/server/middleware/gil.go
@@ -10,15 +10,10 @@ import (
 var lock sync.Mutex
 
 func GIL(c *gin.Context) {
-	locked := lock.TryLock()
-
-	if !locked {
-		c.JSON(http.StatusTooManyRequests, "locked by other request")
-		c.Abort()
-		return
-	}
-
+	// Block and wait for lock instead of immediately failing
+	// This prevents 429 errors when requests queue up briefly
+	lock.Lock()
 	defer lock.Unlock()
+	
 	c.Next()
-
 }


zhiyuan8:
@artik0din Thank you for your contrbution to NexaSDK!

zhiyuan8:
@artik0din thank you, this is a solid fix — replacing TryLock() with a blocking Lock() correctly removes false 429s and maintains proper serialization guarantees. 
For future consideration, we might explore adding a timeout-based locking mechanism to handle potential queuing delays more gracefully under heavy load.

RemiliaForever:
This is by design. Currently, the server does not support concurrent processing.

I believe it is better to queue requests on the client side rather than on the server. This approach allows requests that cannot be handled promptly to be correctly discarded during periods of high load.

Queuing on the server side also introduces potential performance issues. For example, when handling requests for multiple models, concurrency may result in frequent loading and unloading of models, leading to inefficiencies.

zhiyuan8 (approved):

artik0din:
@zhiyuan8 Thanks for the approval and the feedback! I completely agree that this is a pragmatic fix for the immediate HTTP 429 issue we were facing.

@RemiliaForever You raise a very valid point about this being a band-aid solution. I understand the concern about moving the blocking logic to the client side. This PR focuses on resolving the immediate concurrency issue, but you're absolutely right that a more robust server-side solution would be ideal.
For future consideration, I think we could explore:

Timeout-based locking (as @zhiyuan8 mentioned) - to handle high-load scenarios more gracefully
Server-side request queuing - to handle concurrent requests properly at the server level
Better model loading/unloading strategies - to reduce the performance impact when handling multiple models

Would it make sense to open a separate issue to track these improvements? I'd be happy to contribute to those discussions or even work on a follow-up PR if the maintainers are interested.
For now, since this PR addresses the immediate problem and has been approved, I propose we merge it and iterate on a more comprehensive solution separately.

zhiyuan8:
@artik0din I have merged, thanks!

PR #855, opened 2025-11-14
Feature/modelscope support

yrk111222:
This PR primarily aims to make it easy to download models from ModelScope. To download a model, users can use the `ms` or `modelscope` parameter with a command like the following:

`./build/nexa pull NexaAIDev/Qwen3-VL-2B-Instruct-GGUF --model-hub ms`

Users can also set the `NEXA_MS_CACHE_DIR` environment variable to specify a custom download directory. Default model path is like `~/.cache/nexa.ai/modelscope_cache/NexaAIDev/Qwen3-VL-2B-Instruct-GGUF`

For model inference, a similar command can be used to load model files from a designated local directory:
```
./build/nexa pull NexaAIDev/Qwen3-VL-2B-Instruct-GGUF \
    --model-hub localfs \
    --local-path /path/to/modeldir
```

For NexaAIDev models on ModelScope, please refer to the following link:

https://modelscope.cn/organization/NexaAIDev

### DIFF
--- a/runner/cmd/nexa-cli/model.go
+++ b/runner/cmd/nexa-cli/model.go
@@ -219,7 +219,7 @@ func pullModel(name string, quant string) error {
 				return fmt.Errorf("local path is required for localfs model hub")
 			}
 			model_hub.SetHub(model_hub.NewLocalFS(localPath))
-		case "modelscope", "ms":
+		case "ms", "modelsocpe":
 			model_hub.SetHub(model_hub.NewModelScope())
 		default:
 			return fmt.Errorf("unknown model hub: %s", modelHub)


RemiliaForever:
Thanks for your contribution! 

I noticed that your PR uses Python for downloading, which might make things a bit tricky for users, and may prevent the progress bar from displaying correctly. 

Considering that ModelScope can also retrieve file lists and download files via HTTP, would it be better to adopt a logic similar to that of HuggingFace?

yrk111222:
@RemiliaForever Thank you so much for the valuable feedback!

You're absolutely right. I will update my PR to reflect this new implementation. It might be a little later, possibly sometime next week.

PR #909, opened 2025-12-05
feat: functioncall demo with NexaAI/OmniNeural-4B

mengshengwu:


mengshengwu:
plz not use squash merge

Issue #866, opened 2025-11-17
Check disk free space before download it


RemiliaForever:

zhiyuan8:
please provide this to GO project

RemiliaForever:
closed in #962

PR #959, opened 2025-12-26
feat/nexa-cli: check disk space before model download

MohamedAliBouhaouala:
The following PR adds a disk space check before downloading a model by introducing the `hasEnoughDiskSpace` function, which verifies that the target path has sufficient free space according to the model manifest size. 
The check is performed at the start of `pullModel` function before initiating any download. 
### Important Notes:
- No safety buffer has been applied at this time. It can be added at the reviewers' discretion. 
- A new dependency has been added `gopsutil/disk` for cross-platform disk usage check.
### ISSUES: [866](https://github.com/NexaAI/nexa-sdk/issues/866)


RemiliaForever:
merged in #962

PR #984, opened 2026-01-12
chore: only create output image for cv inference if there are meaning…

vinovo:
…ful draw of bounding box / mask


vinovo:
> It doesn't seem necessary to split it into two sections; the newly added code is part of the continuous logic.

OK, merged into a single section.

PR #1003, opened 2026-01-16
feat: add warning for single-threaded downloads when NEXA_HFTOKEN is not set

mengshengwu:


runner/internal/model_hub/model_hub_hf.go
```diff
@@ -48,6 +49,7 @@ func (d *HuggingFace) MaxConcurrency() int {
 	if config.Get().HFToken != "" {
 		return 8
 	} else {
+		fmt.Println(render.GetTheme().Warning.Sprintf("WARN: NEXA_HFTOKEN not set - downloads will use single-threaded mode. Set NEXA_HFTOKEN environment variable for faster multi-threaded downloads"))
```
RemiliaForever: `WARN` -> `Warning`

RemiliaForever: It would be better to add a documentation link for how to get HFTOKEN.

PR #1033, opened 2026-02-03
feat: implement resumable downloads for model files

### LINKED ISSUE #803: nexa pull resumable download support

kashyapkale:
This commit adds comprehensive support for resuming interrupted model downloads, significantly improving user experience when downloading large model files.

Key changes:
- Added checkExistingFile() function to validate existing file state before download
  - Detects complete files and skips re-downloading
  - Identifies partial files for resume capability
  - Handles corrupted files (larger than expected) by removing and re-downloading
  - Returns appropriate size indicators for each state

- Enhanced download logic in StartDownload():
  - Checks for existing files before initiating new downloads
  - Resumes downloads from chunk-aligned offsets to ensure data integrity
  - Properly tracks progress for already-downloaded portions
  - Skips complete files entirely, reporting progress immediately

- Removed pre-download cleanup in Store.Pull():
  - Previously removed models before pulling, preventing resume capability
  - Now preserves existing files to enable resumable downloads

- Added comprehensive test coverage:
  - Tests for nonexistent files
  - Tests for complete files
  - Tests for partial files
  - Tests for corrupted files (larger than expected)
  - Tests for empty files

This implementation ensures that interrupted downloads can be resumed seamlessly, reducing bandwidth usage and improving reliability for large model downloads.

Closes #803

### DIFF
--- a/runner/internal/model_hub/model_hub.go
+++ b/runner/internal/model_hub/model_hub.go
@@ -17,6 +17,7 @@ package model_hub
 import (
 	"bytes"
 	"context"
+	"encoding/binary"
 	"fmt"
 	"io"
 	"log/slog"
@@ -25,7 +26,6 @@ import (
 	"path/filepath"
 	"reflect"
 	"slices"
-	"sort"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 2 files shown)


kashyapkale:
@mengshengwu 

Oh my bad, you're absolutely right, using file size to determine resume position doesn't work with concurrent chunked downloads since
  chunks complete out of order, leaving gaps that file size can't account for. That could silently produce corrupted files.                   
                  
  I've reworked the resume logic based on your suggested approach:

  - Chunk-level tracking via a .chunks marker file, each completed chunk ID is recorded in a small JSON file (<filename>.chunks) stored
  alongside the download. The marker also records file_size, chunk_size, and total_chunks so we can detect and discard stale/incompatible
  markers across versions.
  - Resume schedules only incomplete chunks, on restart, we load the marker, skip already-completed chunks, and only queue the remaining
  ones. The old checkExistingFile (file-size-based) function has been removed entirely.
  - Marker cleanup on success, once all chunks finish, the marker files are deleted.
  - Mutex-protected writes, since multiple goroutines complete concurrently, markComplete() is guarded by a sync.Mutex to safely persist
  state after each chunk.

  Edge cases like corrupted marker JSON, mismatched file/chunk sizes between runs, and backwards compatibility with pre-marker downloads are
  also handled (all default to a safe fresh start).

  Thank you for the direction, this is a much more robust approach!

RemiliaForever:
JSON format is not suitable for concurrent writes; multiple threads must use locks to access it, which defeats the purpose of parallel acceleration. 

JSON is also not ideal for marking-type modifications, as small changes can cause write amplification.

Using a binary format, where each byte maintains the state of a corresponding chunk, is a better choice.

kashyapkale:
@RemiliaForever                                                                                                                            
                                                            
Make sense,  JSON serialization forces a mutex on every chunk completion and rewrites the entire file each time.
                                                                                                                                            
  I've reworked the marker to a binary format:              

  - Layout: 24-byte header (NXCK magic + fileSize/chunkSize/totalChunks as little-endian integers) followed by 1 byte per chunk (0x00 = pending, 0x01 = complete)
  - No mutex: each goroutine writes exactly 1 byte via WriteAt (backed by pwrite) at its own offset — no lock contention, no serialization
  - No write amplification: marking a chunk writes 1 byte instead of re-serializing the entire state
  - A single *os.File handle is kept open for the lifetime of the download and properly closed via defer on all exit paths
  - Old JSON markers are detected (bad magic) and safely replaced with a fresh binary file

All existing tests updated + added: on-disk layout verification, race detector validation, bounds checking, old-JSON migration, and concurrent persistence (reload from disk after parallel writes)

PR #1112, opened 2026-06-27
feat(sdk): apply chat template in LLM bench text-prompt path

RemiliaForever:
## What

The VLM path in `geniex-bench` has always wrapped prompts with a chat template via `geniex_vlm_apply_chat_template`, but the LLM `--prompt-file` (text-prompt) path previously fed the raw text directly into `prompt_utf8` without applying a template. This made LLM `--accuracy` output inconsistent with real inference (LLM is templated in the CLI's `infer.go`).

## Changes

- Add `build_llm_prompt` (mirroring `build_vlm_prompt`): run a single user message through `geniex_llm_apply_chat_template` with `add_generation_prompt = true` and `enable_thinking = true`.
- `run_llm` now applies the template before feeding `gin.prompt_utf8` in `--prompt-file` mode, and frees it with `geniex_free` on both the error-exit and normal-completion paths.
- The VLM path (`build_vlm_prompt`) also sets `enable_thinking = true` to align with LLM.
- Update the file-top comment, `run_vlm` comment, and `--prompt-file` help text accordingly.

Behavioral differences remain reasonable: the default random-ids prefill mode (`-p N`) still skips the tokenizer and template — it is purely a prefill-speed test by design. Template/think only takes effect on the `--prompt-file` path.

## Notes

- Only touches the SDK's C bench program; `geniex.h` is untouched, so this is not an FFI change.
- Builds cleanly locally with `cmake --build build --target geniex-bench`.

### DIFF
--- a/sdk/benchmark/benchmark.c
+++ b/sdk/benchmark/benchmark.c
@@ -54,9 +54,10 @@
 #include <dirent.h>
 #endif
 
-/* Fixed VLM prompt. Multimodal cells still go through the bundle's chat
- * template (which needs real text), so the LLM/VLM paths are intentionally
- * asymmetric: LLM prefills random ids; VLM keeps a one-line text payload. */
+/* Fixed VLM prompt. Multimodal cells always go through the bundle's chat
+ * template (which needs real text); the LLM text-prompt path (--prompt-file)
+ * does the same. LLM diverges only in its default random-ids prefill mode,
+ * which skips the tokenizer and template entirely. */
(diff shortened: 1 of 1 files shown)


zhiyuan8:
@mengshengwu  need to take a look, do not merge till next week after we finish perf scorecard run.

mengshengwu (approved):

mengshengwu:
LGTM

RemiliaForever:
since qairt need directly pass prompt to generate, consider adding a flag.

PR #1119, opened 2026-06-28
fix(build): bundle Android NDK libomp.so into the SDK package

### LINKED ISSUE #1118: [release] Bundle libomp.so into geniex-bench-android-arm64 tar.gz
## Background

`geniex-bench-android-arm64-*.tar.gz` ships without `libomp.so`, yet `libggml-cpu.so` (llama.cpp CPU backend) links OpenMP dynamically and crashes at decode time without it. The Windows/Linux bench archives are already self-contained because the SDK install layer drops the OpenMP runtime into `lib/llama_cpp/` (`sdk/plugins/CMakeLists.txt:9-41` — `libomp140.aarch64.dll` for Windows, `libgomp.so.1` for Linux), which `release.yml`'s platform-agnostic `cp -r .../lib/llama_cpp` then carries into the archive. There is no equivalent Android branch, so the QDC scorecard has to hand-inject `libomp.so` from `bindings/android/app/extLibs/arm64-v8a/` at artifact-build time (`sdk/benchmark/qdc/run_qdc_jobs.py:212-223`).

## Goal

Add the missing Android branch in the SDK install layer so `libomp.so` is sourced from the NDK and installed into `lib/llama_cpp/`, making the Android tar.gz self-contained like the Windows zip. Then drop the now-redundant manual injection in the QDC harness so `pkg-geniex` is the single source.

mengshengwu:
Closes #1118

`geniex-bench-android-arm64-*.tar.gz` shipped without `libomp.so`, but `libggml-cpu.so` links OpenMP dynamically and crashes at decode without it. Windows/Linux bench archives are already self-contained — the SDK install layer drops their OpenMP runtime into `lib/llama_cpp/`, which `release.yml`'s platform-agnostic `cp -r .../lib/llama_cpp` carries into the archive. Android just lacked the matching install branch, so the QDC harness hand-injected `libomp.so` from `bindings/android/app/extLibs/`.

This adds the Android branch in `sdk/plugins/CMakeLists.txt` (sources `libomp.so` from the NDK, installs to `lib/llama_cpp/`) and drops the now-redundant injection in `run_qdc_jobs.py`, making `pkg-geniex` the single source.

## Test plan

- [x] In the `geniex-toolchain-android:v0.0.1` container, the new `file(GLOB_RECURSE)` + `list(FILTER)` resolves `$ANDROID_NDK_ROOT/.../lib/clang/19/lib/linux/aarch64/libomp.so`; `readelf -h` confirms ELF64 / AArch64. (First attempt's single-`*` glob missed the `lib/clang/<ver>/` segment and tripped the `FATAL_ERROR` — fixed.)
- [x] `build-sdk / android-arm64` runs `cmake --install` green on the `v0.3.13-alpha.2` release run; since the Android branch `FATAL_ERROR`s when libomp.so is unresolved, a green install confirms it is installed into `lib/llama_cpp/`. (`release.yml`'s existing `cp -r .../lib/llama_cpp` then carries it into the tar.gz, same as Windows/Linux.)
- [x] QDC Android scorecard cell decodes without an OpenMP-missing crash using the self-contained tar.gz (no `extLibs` injection) — pending, needs device

### DIFF
--- a/sdk/benchmark/qdc/run_qdc_jobs.py
+++ b/sdk/benchmark/qdc/run_qdc_jobs.py
@@ -209,18 +209,6 @@ def build_android_artifact(
     # on-device; results land in the device's QDC_logs and are auto-collected.
     stage = tmp / "stage"
     shutil.copytree(pkg_dir, stage / "pkg-geniex")
-    # libggml-cpu.so needs libomp.so, which the CLI package doesn't ship but the
-    # Android app provides from extLibs; drop it beside the ggml libs.
-    omp = (
-        HERE.parents[2]
-        / "bindings"
-        / "android"
-        / "app"
-        / "extLibs"
-        / "arm64-v8a"
-        / "libomp.so"
-    )
-    shutil.copy(omp, stage / "pkg-geniex" / "lib" / "llama_cpp" / "libomp.so")
     (stage / "matrix_rows.txt").write_text("\n".join(model_rows(models, device)))
     (stage / "chipset.txt").write_text(CHIPSET.get(device, ""))
     shutil.copytree(HERE / "tests", stage / "tests")
--- a/sdk/plugins/CMakeLists.txt
+++ b/sdk/plugins/CMakeLists.txt
@@ -39,3 +39,14 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
             DESTINATION lib/llama_cpp
             RENAME "libgomp.so.1")
 endif()
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Android")
+    file(GLOB_RECURSE _android_libomp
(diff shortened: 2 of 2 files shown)


mengshengwu:
## Test results

**Summary:** the `android-arm64` SDK install tree from the `v0.3.13-alpha.2` release run carries `libomp.so` beside `libggml-cpu.so` — the tar.gz is now self-contained.

Downloaded the `sdk-android-arm64` artifact from run [28322044556](https://github.com/qualcomm/GenieX/actions/runs/28322044556) and inspected the install tree:

`file lib/llama_cpp/libomp.so` → `ELF 64-bit LSB shared object, ARM aarch64, ... built by NDK r27-beta1`

| `lib/llama_cpp/` | size |
|---|---|
| `libggml-cpu.so` | 869768 |
| `libomp.so` | 1233712 |

PR #1128, opened 2026-07-01
fix(build): bump Hexagon toolchain to v0.7 to fix miscompiled HTP v75…

zhycheng614:
… kernel

The linux SDK toolchain image was built FROM snapdragon-toolchain/arm64-linux:v0.1 (Hexagon SDK 6.4.0.2), which miscompiles libggml-htp-v75.so. On QCS8275 (which loads the v75 DSP kernel) this produces gibberish GGUF output on the NPU, while CPU/GPU paths and QAIRT are unaffected. Rebuilding the v75 kernel with the v0.7 toolchain (Hexagon SDK 6.6.0.0) yields correct output, verified by swapping the kernel into the shipped qualcomm/geniex image.

- toolchain-linux.Dockerfile: UPSTREAM_TAG v0.1 -> v0.7
- _build-sdk.yml: pin geniex-toolchain-linux:v0.1.0 (rebuilt derived image)

### DIFF
--- a/.github/docker/toolchain-android.Dockerfile
+++ b/.github/docker/toolchain-android.Dockerfile
@@ -5,16 +5,16 @@
 # bakes in build-essential (for Rust proc-macro host linking), ccache,
 # and the aarch64-linux-android Rust target so _build-sdk.yml can skip
 # the per-run apt-get / rustup installation.
-ARG UPSTREAM_TAG=v0.3
+ARG UPSTREAM_TAG=v0.7
 FROM ghcr.io/snapdragon-toolchain/arm64-android:${UPSTREAM_TAG}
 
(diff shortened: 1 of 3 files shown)


zhycheng614:
CI status notes:

- **build-sdk / linux-arm64** — fails because the pinned container `geniex-toolchain-linux:v0.1.0` is not published yet. The pin bump alone is not enough; the derived image must be rebuilt from the updated `toolchain-linux.Dockerfile` (`UPSTREAM_TAG=v0.7`) and pushed as `v0.1.0` (via the *Build Toolchain Docker* workflow) before this job can pass.
- **Repolinter** — pre-existing failure (missing SPDX headers in `bindings/…` and `tests/…`); not touched by this PR.

PR #1132, opened 2026-07-01
fix(python): scope VLM modality check to the current turn

### LINKED ISSUE #1131: python: multi-turn VLM raises on generate() when prior turn had media
## Background

`GenieXVLM._apply_chat_template` calls `_messages_have_modality(messages, ...)` over the **entire** history (bindings/python/geniex/modeling.py:424-425 + :332-342), then `generate()` refuses to run if `images=[]` / `audios=[]` while the flag is set (bindings/python/geniex/modeling.py:470-478).

This makes any second turn of a VLM chat that started with an image fail, because the history still carries the previous turn's `{"type": "image", ...}` content block:

```
Turn 1: user=[image, "look"] → images=[cat.png] → OK
Turn 2: user="what color?"    → history still holds turn-1 image block
                              → generate(images=[]) raises
                                "messages reference image content but generate(images=[...]) is empty"
```

Go CLI + server keep multi-turn working because they extract media from only the **last** message (cli/server/handler/chat.go:465-474, cli/cmd/geniex/infer.go:502-504).

## Goal

Align Python with Go: `_messages_have_modality` inspects only the last message, so the pre-generate check fires only when the current turn references media without a matching path.

mengshengwu:
Closes #1131

## Summary

`_messages_have_modality` walked the full chat history when deciding whether the pre-`generate()` guard should fire; it now inspects only the last message, matching how the Go CLI / server extract media (only from the current turn).

## Test plan

- [x] `_messages_have_modality` returns `False` when only *prior* turns hold `{type: image}` / `{type: audio}` blocks and the current turn is text-only (new `test_messages_have_modality_ignores_prior_turns`)
- [x] `_messages_have_modality` still fires on the current turn's media block (new `test_messages_have_modality_detects_current_turn`)
- [x] `_messages_have_modality([], modality)` returns `False` (new `test_messages_have_modality_empty`)
- [x] `pytest bindings/python/tests/unit/test_error_messages.py -q` → 25 passed

### DIFF
--- a/bindings/python/geniex/modeling.py
+++ b/bindings/python/geniex/modeling.py
@@ -330,15 +330,21 @@ class GenieXLLM:
 
 
 def _messages_have_modality(messages: list[dict], modality: str) -> bool:
-    # True if any message's content list contains a block with the given
-    # ``type`` (e.g. 'image' / 'audio'). Plain string content trivially
-    # holds none.
-    for msg in messages:
-        content = msg.get('content')
-        if isinstance(content, list):
-            for item in content:
-                if isinstance(item, dict) and item.get('type') == modality:
-                    return True
+    # True if the LAST message's content list contains a block with the
+    # given ``type`` (e.g. 'image' / 'audio'). Only the last turn matters
+    # here: generate(images=[...]) supplies the paths for the current
+    # turn, and stale image/audio blocks in prior turns must not gate it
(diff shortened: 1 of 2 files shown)


mengshengwu:
Superseded by #1130 which fixes the same bug plus a related test ordering issue. Regression tests from this PR have been ported over.

PR #1160, opened 2026-07-07
feat(qairt): opt-in sliding window CLI flag and env var

### LINKED ISSUE #1140: qairt runtime: CJK detokenizer output is mojibake (e.g. 中国 → ä¸ŃåĽ½)
Describe the bug

  On the qairt (NPU) runtime, some CJK output is corrupted into Latin Extended-A/B characters (e.g. 中国 → ä¸ŃåĽ½). In our tests, English output was clean. The issue reproduces across
  multiple different models we tried, suggesting it lives in the detokenizer / string-handling layer rather than in a specific quantization or model.

  To Reproduce

  Repro 1 — NPU runtime, Qualcomm-quantized VLM:

  > geniex.exe --compute npu infer "local/qwen3-vl-qairt"
  > 请说明图片<path-to-image>.jpg的大类别，小类别和是否属于屏幕截图（是/否）。请严格分三行输
  大类别：å®¶åºŃ活动
  小类别：儿童打扫卫生
  是否截图：是

  Repro 2 — interactive CLI, plain Chinese prompt:

  > geniex infer qualcomm/Qwen2.5-VL-7B-Instruct -p "请用一句话介绍北京。" --enable-json
  > 请用一句话介绍北京。
  北京，ä¸ŃåĽ½首都，拥有丰富的历史和文化，以及现代都市的生活节奏。

  Repro 3 — HTTP endpoint (/v1/chat/completions) with a byte-clean UTF-8 request (curl --data-binary @body.json):

  Request body: {"model":"qualcomm/Qwen2.5-VL-7B-Instruct",
                 "messages":[{"role":"user","content":"请用一句话介绍北京。"}],
                 "max_tokens":128}
  Response content: "北京，ä¸ŃåĽ½首都，拥有丰富的历史和文化，以及现代都市的生活节奏。"

  Expected behavior

  Chinese output should render as valid CJK, e.g. 北京，中国首都，.... Other inference stacks we've used (llama.cpp standalone, vLLM, HuggingFace transformers) produce clean CJK output for
  the same model families.

  What we've ruled out

  - Terminal / codepage. chcp 65001 + [Console]::OutputEncoding = UTF8 + $OutputEncoding = UTF8 do not help. Redirected stdout is UTF-16 LE with the mojibake already stored as wide
  characters — corruption happens inside the GenieX process.
  - CLI display layer. HTTP endpoint produces the exact same corrupted string as the CLI. --enable-json and plain output paths are identical.
  - Client-side encoding. Request body verified byte-clean UTF-8 (\xe8\xaf\xb7... for "请") via curl --data-binary.
  - Model / quantization. Reproduces across multiple different Qualcomm-quantized models we tried on qairt.
  - Sampling / template / KV-cache. Reproduces at temperature=0, short single-turn prompts, no KV-cache pressure.
  - Global Unicode issue. Appears to be token-ID specific: in our tests, Japanese Kanji output was clean even though Japanese Kanji and Chinese Hanzi share Unicode code-point ranges.
  Korean was partially affected in our samples (e.g. 국 broke, others rendered fine). This points at a byte-concatenation path over specific vocab-range token IDs, not something that
  touches all CJK code points uniformly.

  Mojibake fingerprint

  All corrupted characters we observed fall in Latin Extended-A/B (U+0100–U+024F). The pattern (ä¸Ń for byte sequence E4 B8 AD = "中") is consistent with UTF-8 bytes being decoded as a
  single-byte Eastern-European codepage (CP1250 / CP1257 family). Most likely culprit on Windows: MultiByteToWideChar called with the system default codepage instead of CP_UTF8, or a
  std::string→std::wstring conversion that assumes single-byte encoding.

  Environment

  - OS: Windows 11 28000
  - GenieX: current main (commit 3e80f9c8)
  - Runtime affected: qairt (NPU)
  - Models observed affected: qualcomm/Qwen2.5-VL-7B-Instruct, local/qwen3-vl-qairt
  - Invocation paths affected: CLI (geniex infer), HTTP endpoint (/v1/chat/completions)

Davidqian123:
## Summary

Wires the opt-in sliding-window context eviction (`geniex-qairt-plugin` PR [#13](https://github.com/qualcomm/geniex-qairt-plugin/pull/13)) into the public SDK API, the CLI, and every language binding, per qcom-ai-hub/geniex#1197.

- Bumps the `third-party/geniex-qairt` submodule pointer.
- Adds `sliding_window` / `sliding_window_n_keep` fields to `geniex_GenerationConfig` (`sdk/include/geniex.h`), read directly by `QairtLlm::generate()`. `llama_cpp` ignores these fields (it always context-shifts).
- Adds a `--sliding-window` flag to `geniex infer`, passed through the typed config on every `Generate()` call.
- Updates every binding's FFI surface in this change (per `CONTRIBUTING.md`): Go (`bindings/go`), Python (`bindings/python`), Android (`bindings/android`).
- Documents the flag in `cli/README.md` and `docs/{en,cn}/run/cli/reference.mdx`.

## Why a typed config field instead of an env var

An earlier version of this PR read an env var (`GENIEX_SLIDING_WINDOW=1`, set via `os.Setenv()` from the Go CLI) directly in the plugin. This had a real bug on Windows: `os.Setenv()` called at runtime from Go is not guaranteed to be visible to a separately-CRT-linked plugin DLL (each DLL's CRT can hold its own environment-block snapshot) -- `--sliding-window` was silently non-functional unless the env var was also exported in the parent shell *before* launching the process.

Moving the flag onto `geniex_GenerationConfig` fixes this (same in-process struct pointer already used for `max_tokens`/`sampler_config`) and, per review feedback, makes the feature available to the Python and Android bindings as well, not just the Go CLI.

## Submodule bump also picks up an EOS-detection fix and a CJK detokenizer fix

The latest `third-party/geniex-qairt` submodule bump in this branch also picks up two unrelated fixes from that repo's history:

- **EOS-detection fix** (geniex-qairt-plugin@a723e23): PR #12 (graph-info refactor) introduced `LLMModel(buildSpecSkeleton(gc), std::move(gc))` in every plain-LLM model factory -- an unsequenced pair of arguments that could leave `eos_token_ids` empty depending on evaluation order, causing generation to never stop at EOS and instead always run to `max_tokens`. See geniex-qairt-plugin#13 for the full root cause and fix; verified on real hardware via the Go CLI that generation now stops cleanly at EOS again.
- **CJK detokenizer fix**: the submodule's own `third-party/geniex-proc` submodule was bumped past PR [qualcomm/geniex-proc#10](https://github.com/qualcomm/geniex-proc/pull/10) ("fix: extend byte-level reverse table to U+0143 to fix CJK detok"), which fixes #1140 -- CJK output on the qairt (NPU) runtime being corrupted into mojibake (e.g. `中国` → `ä¸ŃåĽ½`). Root cause was a byte-level fallback reverse-table that only covered up to a certain code point; CJK byte-fallback tokens beyond that range detokenized incorrectly.

## Acceptance criteria (qcom-ai-hub/geniex#1197)

- [x] Default behavior is preserved: exceeding the context length without opt-in returns the existing out-of-context error.
- [x] Passing `--sliding-window` to the CLI enables ring-buffer eviction.
- [x] With sliding window enabled, conversations continue past the context limit by evicting the oldest context -- verified end-to-end on real QAIRT hardware (see geniex-qairt-plugin#13), including the case where eviction fires mid-turn and the conversation continues coherently into the next turn.
- [x] Available to all bindings (Go, Python, Android), not just the CLI.
- [x] Document the flag in CLI help / docs.

## Testing

- Rebuilt the SDK bridge + qairt plugin (`cmake --build`) and the Go CLI (`bazelisk build //cli --config=windows_arm64`) -- clean builds.
- `clang-format`, `ruff check`/`ruff format --check` on all touched files -- clean.
- Ran `geniex infer qualcomm/Qwen3-4B-Instruct-2507 --sliding-window` on real hardware with `GENIEX_SLIDING_WINDOW` unset from the environment (confirming the old env-var dependency is gone): eviction fires (`sliding window: discarding N tokens...`), generation continues with no `context length exceeded` error, and the model stops cleanly at EOS on both short and long turns.

## Related

- qualcomm/geniex-qairt-plugin#13 (core re-prefill eviction implementation + EOS-detection fix)
- Fixes #1140 (CJK detokenizer mojibake on the qairt runtime), via qualcomm/geniex-proc#10

### DIFF
--- a/bindings/android/app/src/main/cpp/jniutils.cpp
+++ b/bindings/android/app/src/main/cpp/jniutils.cpp
@@ -152,6 +152,13 @@ geniex_GenerationConfig extract_generation_config(JNIEnv* env, jobject configObj
     cfg.audio_paths = audioPathPtrs.empty() ? nullptr : (geniex_Path*)audioPathPtrs.data();
     cfg.audio_count = audioPathPtrs.size();
 
+    // ----------- sliding window (qcom-ai-hub/geniex#1197) -----------
+    jfieldID slidingWindowId = env->GetFieldID(cls, "slidingWindow", "Z");
+    cfg.sliding_window       = slidingWindowId ? env->GetBooleanField(configObj, slidingWindowId) : false;
+
+    jfieldID slidingWindowNKeepId = env->GetFieldID(cls, "slidingWindowNKeep", "I");
+    cfg.sliding_window_n_keep     = slidingWindowNKeepId ? env->GetIntField(configObj, slidingWindowNKeepId) : 0;
+
     return cfg;
 }
 
--- a/bindings/android/app/src/main/java/com/geniex/sdk/bean/GenerationConfig.kt
+++ b/bindings/android/app/src/main/java/com/geniex/sdk/bean/GenerationConfig.kt
@@ -10,5 +10,9 @@ data class GenerationConfig(
     var imagePaths: Array<String>? = null,
     var imageCount: Int = 0,
     var audioPaths: Array<String>? = null,
-    var audioCount: Int = 0
+    var audioCount: Int = 0,
+
+    // Opt-in ring-buffer context eviction (qairt only).
+    var slidingWindow: Boolean = false,
+    var slidingWindowNKeep: Int = 0 // 0 = plugin default (4)
 )
--- a/bindings/go/common.go
+++ b/bindings/go/common.go
@@ -104,15 +104,21 @@ type GenerationConfig struct {
 	ImagePaths     []string
 	ImageMaxLength int32
 	AudioPaths     []string
+
+	// Opt-in ring-buffer context eviction (qairt only).
+	SlidingWindow      bool
+	SlidingWindowNKeep int32 // 0 = plugin default (4)
 }
 
 // LCOV_EXCL_START
 func (gc GenerationConfig) toCPtr() *C.geniex_GenerationConfig {
 	cPtr := (*C.geniex_GenerationConfig)(cMalloc(C.sizeof_geniex_GenerationConfig))
 	*cPtr = C.geniex_GenerationConfig{
-		max_tokens:       C.int32_t(gc.MaxTokens),
-		n_past:           C.int32_t(gc.NPast),
-		image_max_length: C.int32_t(gc.ImageMaxLength),
+		max_tokens:            C.int32_t(gc.MaxTokens),
+		n_past:                C.int32_t(gc.NPast),
+		image_max_length:      C.int32_t(gc.ImageMaxLength),
+		sliding_window:        C.bool(gc.SlidingWindow),
+		sliding_window_n_keep: C.int32_t(gc.SlidingWindowNKeep),
 	}
 
 	if len(gc.Stop) > 0 {
--- a/bindings/python/geniex/_ffi/_types.py
+++ b/bindings/python/geniex/_ffi/_types.py
@@ -82,6 +82,8 @@ class geniex_GenerationConfig(Structure):
         ('image_max_length', c_int32),
         ('audio_paths', POINTER(c_char_p)),
         ('audio_count', c_int32),
+        ('sliding_window', c_bool),
+        ('sliding_window_n_keep', c_int32),
     ]
 
 
--- a/bindings/python/geniex/generation/config.py
+++ b/bindings/python/geniex/generation/config.py
@@ -29,6 +29,6 @@ class GenerationConfig:
     audios: list[str] = field(default_factory=list)
     stream: bool = False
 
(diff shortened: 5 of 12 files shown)


mengshengwu:
It looks good, but I have a few questions:
1. Why pass this parameter via environment variables? Since it is within gen_cfg, wouldn't adding a new struct field be a better approach?
2. Only the Go CLI is supported at present. We have two other client sides, Python and Android. Should we implement this feature for them as well? Taking this into consideration, I believe the solution mentioned in the first point is more appropriate.

Davidqian123:
@mengshengwu Good catch, already Implemented.

mengshengwu (approved):

alexchen4ai:
LGTM.

PR #1163, opened 2026-07-08
feat(sdk): load custom QNN library via GENIEX_QNN_LIB / --qnn-lib

zhycheng614:
## Summary

Adds the ability to point a single GenieX release at a different QAIRT/QNN library build at runtime, so users and internal testers can validate multiple QAIRT versions without rebuilding or reinstalling.

- **`GENIEX_QNN_LIB=<dir>`** — env var read by the qairt plugin. Points at a folder that directly contains the QNN shared libraries (`QnnHtp.dll` / `QnnSystem.dll` / `QnnHtpNetRunExtensions.dll`, or the `libQnn*.so` equivalents).
- **`--qnn-lib <dir>`** — CLI convenience flag that exports `GENIEX_QNN_LIB` for the process. The flag wins when both are set.
- **Default unchanged** — when neither is set, the plugin uses the bundled QNN-lib exactly as before.
- **Fails fast** — a missing directory or one lacking the backend library throws at model load with a clear message (e.g. `GENIEX_QNN_LIB does not contain QnnHtp.dll: <path>`), surfaced via the existing `geniex_llm_create` error path.

### Design notes

- No FFI/ABI change: `geniex.h` is untouched, so no binding FFI updates are required (CONTRIBUTING §4 N/A). Because the mechanism is an env var read inside the plugin, **all** front-ends (CLI, pybind, Android/JNI) pick it up for free.
- No submodule change: `QnnRuntimeConfig` in `third-party/geniex-qairt` already exposes `backend_path`/`system_lib_path`/`extensions_path` overrides. The change only populates them from a new source inside the SDK wrapper (`sdk/plugins/qairt`), which is main-repo code.
- Lib-name literals were de-duplicated into `kQnnBackendLib`/`kQnnSystemLib`/`kQnnExtensionsLib` constants shared by the override and bundled paths.

## Test plan

- [ ] Build the SDK bridge (`/build`) and CLI on a Snapdragon Windows host.
- [ ] `geniex infer local/granite4_micro --qnn-lib <alt-qnn-libs>` loads the alternate libraries (verify via `GENIEX_LOG=info` line "Using custom QNN libraries from GENIEX_QNN_LIB").
- [ ] `GENIEX_QNN_LIB=<alt-qnn-libs> geniex infer local/granite4_micro` behaves identically to the flag.
- [ ] A nonexistent path and a valid-but-empty folder each fail at load with the expected error message.
- [ ] No flag / no env var → still loads the bundled QNN-lib and produces correct output (default unchanged).
- [ ] Swap two different QAIRT/QNN versions against the same GenieX build and confirm each runs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cli/cmd/geniex/infer.go
+++ b/cli/cmd/geniex/infer.go
@@ -76,7 +76,7 @@ var (
 		llmFlags := pflag.NewFlagSet("LLM/VLM Model", pflag.ExitOnError)
 		llmFlags.SortFlags = false
 		llmFlags.StringVarP(&computeUnit, "compute", "c", "", "compute unit to run on: cpu, gpu, npu, or hybrid (default: hybrid for llama_cpp, npu for qairt)")
-		llmFlags.StringVarP(&qnnLib, "qnn-lib", "", "", "path to a folder holding custom QNN libraries (qairt only; overrides GENIEX_QNN_LIB, default: bundled QNN-lib)")
+		llmFlags.StringVarP(&qnnLib, "qnn-lib", "", "", "QAIRT SDK root or lib folder for testing a custom QNN build (qairt only; sets GENIEX_QNN_LIB; may crash on ABI-incompatible QAIRT versions; default: bundled QNN-lib)")
 		llmFlags.Int32VarP(&ngl, "ngl", "n", 0, "number of layers to offload to gpu/npu (llama_cpp only, default 999)")
 		llmFlags.Int32VarP(&nctx, "nctx", "", 0, "context window size (llama_cpp only, default 4096)")
 		llmFlags.Int32VarP(&maxTokens, "max-tokens", "", 2048, "max tokens")
--- a/cli/cmd/geniex/infer.go
+++ b/cli/cmd/geniex/infer.go
@@ -38,6 +38,7 @@ var (
 	input          string
 	systemPrompt   string
 	computeUnit    string
+	qnnLib         string
 
 	// sampler config
 	temperature       float32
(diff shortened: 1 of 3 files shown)


Davidqian123:
I think the main reason we need to support this feature is that we need to integrate geniex into workbench and then workbench can run geniex in different QAIRT SDK versions. Therefore, it is important for us to consider library file structres in the QAIRT SDK folder, because it will not out all lib files we need to run geniex in the same folder. I think we can refer to [this doc](https://github.com/qualcomm/ai-hub-apps/tree/main/tutorials/llm_on_genie#windows-powershell) to set up environment variables when implementing this feature

alexchen4ai:
> I think the main reason we need to support this feature is that we need to integrate geniex into workbench and then workbench can run geniex in different QAIRT SDK versions. Therefore, it is important for us to consider library file structres in the QAIRT SDK folder, because it will not out all lib files we need to run geniex in the same folder. I think we can refer to [this doc](https://github.com/qualcomm/ai-hub-apps/tree/main/tutorials/llm_on_genie#windows-powershell) to set up environment variables when implementing this feature

Yes, correct. Can you apply necessary change here. @Davidqian123 @zhycheng614

zhycheng614:
> > I think the main reason we need to support this feature is that we need to integrate geniex into workbench and then workbench can run geniex in different QAIRT SDK versions. Therefore, it is important for us to consider library file structres in the QAIRT SDK folder, because it will not out all lib files we need to run geniex in the same folder. I think we can refer to [this doc](https://github.com/qualcomm/ai-hub-apps/tree/main/tutorials/llm_on_genie#windows-powershell) to set up environment variables when implementing this feature
> 
> Yes, correct. Can you apply necessary change here. David Qian (@Davidqian123) Perry Cheng (@zhycheng614)

OK, I will fix this.

zhycheng614:
Good call @Davidqian123 — you're right that a real QAIRT SDK doesn't put everything in one flat folder. Pushed c2e6cd8a to handle the actual SDK layout, following the [llm_on_genie tutorial](https://github.com/qualcomm/ai-hub-apps/tree/main/tutorials/llm_on_genie#windows-powershell):

`GENIEX_QNN_LIB` (and `--qnn-lib`) now accepts a **QAIRT SDK root**. The plugin resolves the two lib locations separately, matching how the tutorial sets the env vars:

- **Host libs** (`QnnHtp.dll` / `QnnSystem.dll` / `QnnHtpNetRunExtensions.dll`) from `lib/<triple>` — `aarch64-windows-msvc`, `aarch64-android`, or `aarch64-oe-linux-gcc11.2`. On Windows this dir is also fed to `SetDllDirectoryA` so the transitive stub DLLs resolve.
- **`ADSP_LIBRARY_PATH`** is pointed at *every* `lib/hexagon-v*/unsigned` folder (semicolon/colon-joined), so QNN's loader picks the skel matching the on-device HTP arch — no need to hardcode v73/v75/etc.

A flat folder (our bundled `htp-files` shape) still works unchanged, and there's a `lib/*` fallback scan for triples that vary across SDK versions. Bad paths still fail fast with a clear message that now names both lookup locations.

Docs in `notes/run.md` updated with both layouts. Since this can't build/run in my env, the QAIRT-SDK-root swap still needs an on-device smoke test (test plan in the PR body).

zhycheng614:
Will build in [AIHW/workbench-build](https://github.com/qualcomm/GenieX/tree/AIHW/workbench-build)

PR #1165, opened 2026-07-09
feat(sdk): pull models from Docker Hub's ai/* namespace

ericcurtin:
## Summary

Adds support for pulling GGUF model artifacts published under
[hub.docker.com/u/ai](https://hub.docker.com/u/ai) (e.g. `ai/gemma3`,
`ai/smollm2`), the same protocol `docker model pull ai/<repo>` uses.
The example implementation this follows is
[`docker/model-runner`](https://github.com/docker/model-runner)'s
`pkg/distribution` package.

- **`sdk/model-manager/crates/core/src/source/dockerhub.rs`** (new): a
  `ModelSource` that talks the plain Docker Registry HTTP API V2 —
  anonymous bearer token exchange, a content-negotiated manifest GET,
  and a small config blob GET (carries the GGUF quantization tag and
  tells us whether the model is a VLM via an `mmproj` layer) — all
  before any weight bytes stream, fitting the existing
  plan-then-fetch contract the HuggingFace and AI Hub sources already
  use.
  - Supports the Docker proprietary format only: "v0.1"
    (single-layer-per-format, positional naming — `model.gguf` or
    `model-%05d-of-%05d.gguf` shards, matching
    `docker/model-runner`'s `unpackLegacy` exactly) and "v0.2"
    (layer-per-file with `org.cncf.model.filepath` annotations).
  - Rejects safetensors/DDUF-only images and CNCF ModelPack artifacts
    with an actionable error instead of mishandling them (not yet
    supported).
- **`mapping.rs`**: `is_docker_hub_reference` / `docker_hub_repo_from_name`
  so `docker.io/ai/gemma3` (or `index.docker.io/…`,
  `https://hub.docker.com/r/…`) auto-routes to Docker Hub without an
  explicit hub flag. A bare `ai/gemma3` is deliberately **not**
  auto-routed — that shape is indistinguishable from a HuggingFace
  repo of the same name, so it needs `--model-hub docker` (mirrors
  how HF/AI Hub ambiguity is already handled for `qualcomm/*`).
- **`pull.rs`**: new `PullIntent::DockerHub { repo, reference }`.
- **FFI (`crates/ffi/src/pull.rs`, `include/geniex_model.h`)**: new
  `GENIEX_HUB_DOCKER` hub selector. The existing `quant` field doubles
  as the Docker tag/digest for this hub (empty → `"latest"`), since
  there's no GGUF-quant-filter step to reuse it for otherwise.
- **Go bindings + CLI**: `--model-hub docker`, plus
  `SplitNamePrecisionCaseSensitive` / `IsDockerHubReference` so a
  Docker tag isn't mistakenly upper-cased the way a GGUF
  quantization label is (`ai/gemma3:latest` must not become
  `:LATEST`).
- **README**: one more quickstart line showing
  `geniex infer docker.io/ai/gemma3`.

## Try it

```bash
geniex pull docker.io/ai/gemma3
geniex infer ai/gemma3 --model-hub docker
geniex infer docker.io/ai/gemma3:F16   # explicit tag
```

## Test plan

- [x] `cargo test --workspace` in `sdk/model-manager` (existing 151 +
      5 executor/proxy/pull-resume integration suites, plus 7 new
      `source::dockerhub` unit tests covering: single-GGUF v0.1
      layout, VLM detection via `mmproj` layer, multi-shard GGUF
      naming, v0.2 filepath-annotation naming, safetensors-only
      rejection, and 404 → `HubModelNotFound`).
- [x] `cargo clippy --workspace --all-targets` — no new warnings.
- [x] `cargo fmt --check` — clean.
- [x] `mapping::` unit tests for `is_docker_hub_reference` /
      `docker_hub_repo_from_name` prefix handling.
- [x] `go build ./...` for `bindings/go` + `cli` — not verified in
      this environment (requires the full CMake-staged
      `sdk/pkg-geniex` header/lib tree from `/build`); changes are
      additive to the existing enum/struct and were reviewed by hand
      + `gofmt`.


alexchen4ai:
@RemiliaForever @mengshengwu Plz help review! Thanks for your support! @ericcurtin

mengshengwu:
Heads-up on the `build-sdk / windows-arm64` failure — it's **not** your code. On a public repo, GitHub doesn't pass repo secrets to a fork's PR workflow, so `HEXAGON_HTP_CERT_PFX` is empty and the HTP-cert step in `_build-sdk.yml` hard-`throw`s before CMake even runs. (Every other build job passed once I approved the run.)

Fix is up in #1171: when the signing secret is absent, the step now warns and builds **unsigned** HTP skels instead of failing — the signing block in llama.cpp's `ggml-hexagon/CMakeLists.txt` is already gated on the cert being present, and Release CI re-signs at ship time, so nothing downstream regresses.

Once #1171 lands, please rebase this PR onto `main` to pick it up (a fork PR runs the workflow from its own head, so it needs the rebase — not just the merge to main). Then `windows-arm64` should go green here.

ericcurtin:
@mengshengwu note the models in docker.io/ai are moving to CNCF ModelPack standard. One of the main values of this standard is people can pull from:

HuggingFace, Qualcomm AI Hub Models or Docker Hub

And mirror models on their internal OCI registries (which around 95% of companies have on their pre-existing infrastructure). Reducing the need to pull models over the public internet.

mengshengwu:
Thanks @ericcurtin, good to know. The internal OCI-mirror use case is compelling for us — we'll keep ModelPack in mind for future updates as docker.io/ai moves over.

mengshengwu (approved):

PR #1209, opened 2026-07-21
feat(sdk): add --delay to geniex-bench for inter-run cooldown

RemiliaForever:
Sleep a configurable number of seconds between runs so the device can cool down between repetitions. Applies to warmup and measured runs; no sleep after the last run. Fractional seconds supported. Recorded in the JSON report's params as delay_s.


mengshengwu (approved):

RemiliaForever:
now scorecard use single process eval single prompt

PR #1222, opened 2026-07-23
fix(server): unblock cross-origin browser clients in CORS middleware

### LINKED ISSUE #1168: OpenAI-Compatible API doesn't function through websites
**Describe the bug**
When using the "geniex serve" command to start a server, it seems to be inaccessible through any website in a web browser other than the in-built "Swagger UI"

**To Reproduce**
1. Start the server ("geniex serve")
2. Go to any website that accepts OpenAI-Compatible APIs (for instance: [this](https://aryan-rajoria.github.io/ai-test-webpage/))
3. Input the proper base URL ("http://127.0.0.1:18181/v1") and a model name (i tried the one from the example python script: "unsloth/Qwen3-4B-GGUF:Q4_0"), and provide a prompt and system prompt
4. See error when you send a request

**Expected behavior**
A response to the prompt

**Screenshots**
(from a different test website)

Image

(from the terminal after trying the test website linked before)

Image

**Desktop (please complete the following information):**
 - OS: Windows 11 25H2
 - Browser Zen (Firefox) [Also tested on Edge (Chromium), same result]
 - Version Latest (v0.3.14)

**Additional context**
The example Python script provided in the docs seems to work perfectly. I also tested just llama.cpp and the website was able to access that correctly. Let me know if there's any more info I need to send, I'm unfamiliar with this kind of thing so this could also be user error on my part.

mansiverma897993:
## Problem

`geniex serve` was unreachable from any cross-origin web page (Open WebUI, hosted OpenAI-compatible frontends), while the built-in Swagger UI and non-browser clients (Python, curl) worked fine.

## Cause

Two issues in `cli/server/middleware/cors.go`:

1. The middleware sent `Access-Control-Allow-Credentials: true` together with the default wildcard `Access-Control-Allow-Origin: *`. The Fetch spec forbids that combination, so browsers rejected every cross-origin response. Same-origin and non-browser clients don't run CORS checks, which is exactly the split reported in the issue.
2. `Access-Control-Allow-Headers` was missing `Authorization`, so any UI sending `Authorization: Bearer <key>` would fail preflight even with the wildcard fixed.

## Fix

Minimal change, as suggested by @mengshengwu in the issue thread:

- Drop the `Access-Control-Allow-Credentials` header — the server has no cookie-based auth, so nothing relied on it.
- Add `Authorization` to `Access-Control-Allow-Headers`.

## Testing

- `go build`, `go vet`, and `go test` on `cli/server/middleware` all pass.

Fixes #1168

### DIFF
--- a/cli/server/middleware/cors.go
+++ b/cli/server/middleware/cors.go
@@ -13,8 +13,7 @@ func CORS(c *gin.Context) {
 	h := c.Writer.Header()
 	h.Set("Access-Control-Allow-Origin", config.Get().Origins)
 	h.Set("Access-Control-Allow-Methods", "OPTIONS, GET, POST")
-	h.Set("Access-Control-Allow-Headers", "Content-Type, GenieX-KeepCache")
-	h.Set("Access-Control-Allow-Credentials", "true")
+	h.Set("Access-Control-Allow-Headers", "Authorization, Content-Type, GenieX-KeepCache")
 	h.Set("Access-Control-Max-Age", "86400")
 
 	if c.Request.Method == "OPTIONS" {


mengshengwu (approved):

mansiverma897993:
@mengshengwu Thankyou so much for approval I enjoying contribution :)

Issue #1220, opened 2026-07-23
v0.3.16 Windows: llama.cpp backend crashes with GGML_ASSERT(device) due to signing issue


mengshengwu:
## Summary

The Windows ARM64 build of v0.3.16 has a code-signing regression that breaks the `llama_cpp` backend. Any `infer` / `serve` call that routes to llama.cpp aborts before the first token.

## Symptom

```
$env:GENIEX_LOG='trace'
geniex infer <any llama.cpp model>
```

Crashes with:

```
ggml-backend.cpp:569: GGML_ASSERT(device) failed
```

## Scope

- Affects: v0.3.16, Windows ARM64, `llama_cpp` backend (all models).
- Not affected: `qairt` backend, other platforms, v0.3.14 and earlier.

## Workaround

Roll back to v0.3.14 while a fixed build is in progress:

https://github.com/qualcomm/GenieX/releases/tag/v0.3.14

## Status

Root-caused to the signing step in the v0.3.16 release pipeline. Working with the signing team on a corrected build. This issue tracks the fix and the release; individual user reports of the same crash on v0.3.16 will be closed as duplicates and pointed here.

Related:
- #1217

mansiverma897993:
@mengshengwu take a look on #1233 and let me know changes !!

RemiliaForever:
should fixed on `v0.3.17`

PR #1233, opened 2026-07-24
fix(ci): verify HTP bundle is WHQL-signed before overlaying

mansiverma897993:
## Problem

Related: #1220.

The `overlay-htp` release job trusts any object it finds at `llama-cpp/libggml-htp-<sha>.zip`: a successful `curl` is unzipped straight over the SDK and reported as `signed=true`, with no check of what was actually downloaded.

I compared the shipped artifacts to confirm this is what bit v0.3.16:

| | `libggml-htp.cat` signer | `libggml-htp.inf` |
|---|---|---|
| v0.3.14 (works) | `CN=Microsoft Windows Hardware Compatibility Publisher` (WHQL) | shipped |
| v0.3.16 (crashes) | `CN=QUALCOMM Inc.` via DigiCert only — the pre-WHQL submission signature | **missing** |

Windows only loads HTP skels whose catalog carries the Microsoft WHQL signature, so the Hexagon backend gets no device and every `llama_cpp` call aborts with `GGML_ASSERT(device)` — while the release still shipped without the `-selfsigned` suffix, defeating the existing gate.

This PR doesn't replace the corrected build the team is preparing — it fixes the pipeline defect so a wrong bucket object can't ship as a "signed" release again.

## Fix

Stage the bundle in a temp dir and verify before overlaying:

- `libggml-htp.cat` and at least one `libggml-htp-v*.so` are present;
- the catalog's PKCS#7 signer chain includes `CN=Microsoft Windows Hardware Compatibility Publisher` — checked with `openssl pkcs7 -print_certs`, available on the ubuntu runner;
- `libggml-htp.inf` is restored from the llama.cpp submodule (already checked out in this job) when the bundle omits it.

On any failure the job warns and keeps the self-signed build, so the artifact gets the `-selfsigned` suffix and the documented stable-tag gate in `notes/release.md` catches it before publish.

## Testing

- Ran the exact verification block against the real shipped catalogs: v0.3.14 → **ACCEPT** (WHQL-signed); v0.3.16 → **REJECT** (`signer: DigiCert / QUALCOMM Inc.`).
- Workflow YAML validated with PyYAML.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -131,13 +131,33 @@ jobs:
           set -euo pipefail
           sha=$(git -C third-party/llama.cpp rev-parse --short=6 HEAD)
           url="https://qaihub-public-assets.s3.us-west-2.amazonaws.com/llama-cpp/libggml-htp-${sha}.zip"
-          if curl -fsSL -o htp-signed.zip "$url"; then
-            unzip -o htp-signed.zip -d sdk-windows-arm64/lib/llama_cpp/
-            signed=true
-          else
-            signed=false
+          signed=false
+          # Stage and verify before touching the SDK: Windows only loads HTP
+          # skels whose catalog carries the Microsoft WHQL signature, so a
+          # wrong object in the bucket (e.g. the pre-WHQL submission zip,
+          # signed by Qualcomm's own cert) must not silently replace the
+          # self-signed build and ship as a "signed" release (#1220).
+          if curl -fsSL -o htp-signed.zip "$url" && unzip -o htp-signed.zip -d htp-staged/; then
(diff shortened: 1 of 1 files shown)


mengshengwu:
Thanks Mansi, the artifact comparison is spot on.

That said, this doesn't need a CI change. The v0.3.16 regression came from our **internal signing-delivery flow** dropping the wrong object at the bucket key — not from a defect in this workflow step. We're fixing it at the source; the CI's existing contract ("trust the bucket object at this key") stays valid, and grep-matching the WHQL subject DN here would be brittle for future compliant signing chains. 
Closing this

mansiverma897993:
@mengshengwu ok fine !! Feel free to assign me any issue in future !!

PR #1246, opened 2026-07-28
fix(server): send spec-compliant finish_reason on stream chunks

mansiverma897993:
Fixes #1243

## Problem

Streaming responses serialized `"finish_reason": ""` on every chunk and never emitted a terminal chunk, so OpenAI-compatible clients (e.g. AnythingLLM) never detected completion.

## Fix

`cli/server/handler/chat.go` now uses small local chunk types with spec-correct JSON: intermediate chunks carry `"finish_reason": null`, and every stream (plain text, tool-call, and the tool-parse fallback) ends with an empty-delta chunk whose `finish_reason` is mapped from the SDK stop reason (`stop` / `length` / `tool_calls`) before `data: [DONE]`. Non-streaming responses are unchanged.

## Test report (`-c cpu`, no Snapdragon device)

Environment: x86-64 Ubuntu 24.04 (WSL2), Intel i5-1235U. SDK built with `cmake --preset default -DGENIEX_PLUGIN_QAIRT=OFF` (llama.cpp CPU backend). Server: `geniex serve -c cpu`, model `Qwen/Qwen3-0.6B-GGUF:Q8_0` (HF).

| # | Scenario | Result |
|---|----------|--------|
| 1 | `stream:true` basic | ✅ intermediates `"finish_reason":null`; terminal `{"delta":{},"finish_reason":"stop"}`; `[DONE]` |
| 2 | `stream_options.include_usage` | ✅ usage chunk (`"choices":[]`) after the finish chunk |
| 3 | `max_completion_tokens:8` | ✅ terminal `{"delta":{},"finish_reason":"length"}` |
| 4 | `stream:false` | ✅ unchanged, `"finish_reason":"stop"` |
| 5 | `stream:true` + `tools` | ✅ tool-call delta, then terminal `{"delta":{},"finish_reason":"tool_calls"}` |

Sample stream tail:

```
data:{"object":"chat.completion.chunk","choices":[{"index":0,"delta":{"role":"assistant","content":"😊"},"finish_reason":null}]}
data:{"object":"chat.completion.chunk","choices":[{"index":0,"delta":{},"finish_reason":"stop"}]}
data:[DONE]
```

`go test ./server/... ./cmd/...` pass. Branch rebased onto latest `main`.

PR #1247, opened 2026-07-28
fix(sdk): resolve QAIRT top-level precision as quant alias

### LINKED ISSUE #1242: serve — /v1/models returns model ID with precision suffix, breaking OpenAI-compatible clients
**Describe the bug**
`/v1/models` endpoint returns model ID with precision suffix (e.g. `qualcomm/Qwen3-4B-Instruct-2507:W4A16`), but using this exact ID in `/v1/chat/completions` fails with `SDKError(Invalid input parameters or handle), quantization 'W4A16' not found for model 'qualcomm/Qwen3-4B-Instruct-2507'`. The request only succeeds when the suffix is omitted. This breaks any OpenAI-compatible client that auto-populates the model ID from `/v1/models` (confirmed with AnythingLLM).

**To Reproduce**
1. Pull a QAIRT model: `geniex pull ai-hub-models/Qwen3-4B-Instruct-2507`
2. Start server: `geniex serve --host 127.0.0.1:18181 --compute npu`
3. Query `/v1/models`:
```powershell
Invoke-RestMethod -Uri "http://127.0.0.1:18181/v1/models" -Method Get
```
Returns `qualcomm/Qwen3-4B-Instruct-2507:W4A16`

4. Use the returned ID in a chat request:
```powershell
Invoke-RestMethod -Uri "http://127.0.0.1:18181/v1/chat/completions" -Method Post -ContentType "application/json" -Body '{"model": "qualcomm/Qwen3-4B-Instruct-2507:W4A16", "messages": [{"role": "user", "content": "Say hello"}]}'
```
5. See error: `SDKError(Invalid input parameters or handle), quantization 'W4A16' not found`

**Expected behavior**
`/v1/models` should return a model ID that works directly with `/v1/chat/completions`. Either the precision suffix should be stripped from the models list, or the completions endpoint should accept it.

**Workaround**
Clear the `Precision` field in `geniex.json`:
```powershell
$json = Get-Content "C:\LocalAI\geniex\models\models\qualcomm\Qwen3-4B-Instruct-2507\geniex.json" | ConvertFrom-Json
$json.Precision = ""
$json | ConvertTo-Json -Depth 10 | Set-Content "C:\LocalAI\geniex\models\models\qualcomm\Qwen3-4B-Instruct-2507\geniex.json"
```

**Desktop:**
- OS: Windows 11 ARM64
- Device: Snapdragon X Plus X1P42100
- GenieX CLI Version: v0.3.17
- QAIRT Runtime Version: v2.45.0.260326
- LlamaCPP Runtime Hash: ae9291e
- Model: `ai-hub-models/Qwen3-4B-Instruct-2507` (W4A16, QAIRT runtime)
- AnythingLLM Version: v1.15.0 (used for OpenAI-compatible client testing)

mansiverma897993:
Fixes #1242

## Root cause

QAIRT (AI Hub) manifests store their single `model_file` entry under the placeholder key `"N/A"` and record the real precision (e.g. `W4A16`) on the manifest's top-level `Precision` field. The two consumers of that manifest disagree:

- **Listing** (`geniex_model_list_detailed`, used by `geniex list` and serve `/v1/models`) deliberately surfaces the top-level precision instead of the `N/A` placeholder — so `/v1/models` advertises `qualcomm/Qwen3-4B-Instruct-2507:W4A16`.
- **Resolution** (`resolve_model_paths`, reached from `/v1/chat/completions` via `ModelGetPaths`) only matched `model_file` keys — `W4A16` isn't one, so the advertised id failed with `quantization 'W4A16' not found`.

Any OpenAI-compatible client that auto-populates the model id from `/v1/models` (AnythingLLM in the report) therefore couldn't chat at all.

## Fix

The issue offered two options — strip the suffix from the listing, or make completions accept it. Stripping would regress the human-meaningful precision display that the listing intentionally surfaces, so this PR takes the second option: `resolve_model_paths` now accepts the manifest's top-level precision (case-insensitively, mirroring the FFI's quant-tag normalization) as an alias for the `"N/A"` entry. Advertised ids round-trip; exact `model_file` keys still win over the alias; unknown quants still error; `QuantNotDownloaded` semantics are preserved.

This also fixes the same round-trip for `geniex run name:W4A16` on the CLI, and makes `/v1/models/{id}` (which already accepted the suffix) consistent with completions.

## Testing

- 5 new unit tests in `crates/core/src/paths.rs`: alias hit, case-insensitive alias, alias-but-not-downloaded, unknown quant still errors, exact key wins over alias.
- Full suites green locally: `cargo test -p model-manager-core -p model-manager-ffi` (171 core tests passed), `cargo fmt --all --check` clean.
- No public header / FFI change — the fix is entirely inside the core resolver.

I don't have Snapdragon hardware to run `geniex serve` end-to-end — @Novatoris could you confirm with AnythingLLM once a patched build is available? (Together with #1246 this should make the AnythingLLM flow fully work.)

### DIFF
--- a/sdk/model-manager/crates/core/src/paths.rs
+++ b/sdk/model-manager/crates/core/src/paths.rs
@@ -18,10 +18,19 @@ pub struct ModelPaths {
     pub model_type: ModelType,
 }
 
+/// Placeholder key QAIRT (AI Hub / localfs) manifests use for their single
+/// `model_file` entry; the human-meaningful precision (e.g. "W4A16") lives on
+/// the manifest's top-level `Precision` field instead.
+pub(crate) const PRECISION_NA: &str = "N/A";
+
 /// Resolve file paths from a manifest + local base directory + optional quant hint.
 ///
 /// Replicates the logic in cli/server/service/keepalive.go:121-141:
 /// - If `quant` is Some, look up that exact key; error if not downloaded.
+///   The manifest's top-level precision is accepted as an alias for the
+///   [`PRECISION_NA`] entry: listings (`geniex list`, serve `/v1/models`)
+///   surface that precision as the id suffix, so the advertised
+///   `name:precision` must resolve back here (#1242).
 /// - If `quant` is None, prefer the highest-ranked entry in
 ///   [`QUANT_PRIORITY`]; fall back to lexicographic min when none of the
 ///   downloaded quants appear in the priority list.
@@ -34,17 +43,24 @@ pub fn resolve_model_paths(
 
     let (resolved_quant, model_path) = {
         let (q, file_info) = if let Some(q) = quant {
-            let fi = manifest
-                .model_file
-                .get(q)
-                .ok_or_else(|| Error::QuantNotFound(q.to_string(), manifest.name.clone()))?;
+            let (q, fi) = match manifest.model_file.get(q) {
+                Some(fi) => (q.to_string(), fi),
+                None if !manifest.precision.is_empty()
+                    && q.eq_ignore_ascii_case(&manifest.precision) =>
+                {
+                    let fi = manifest.model_file.get(PRECISION_NA).ok_or_else(|| {
+                        Error::QuantNotFound(q.to_string(), manifest.name.clone())
+                    })?;
+                    (PRECISION_NA.to_string(), fi)
+                }
+                None => {
+                    return Err(Error::QuantNotFound(q.to_string(), manifest.name.clone()));
+                }
+            };
             if !fi.downloaded {
-                return Err(Error::QuantNotDownloaded(
-                    q.to_string(),
-                    manifest.name.clone(),
-                ));
+                return Err(Error::QuantNotDownloaded(q, manifest.name.clone()));
             }
-            (q.to_string(), fi)
+            (q, fi)
         } else {
             let downloaded: Vec<&str> = manifest
                 .model_file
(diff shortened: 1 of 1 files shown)


Novatoris:
@mansiverma897993 Thank you for the quick fix. Could you provide a pre-built Windows ARM64 test binary that includes both this fix (#1247) and #1246? I am happy to test on Snapdragon hardware.

mansiverma897993:
@Novatoris Unfortunately I can't provide a pre-built binary I don't have a Windows ARM64  less resources I have to build environment (the full build needs the Hexagon SDK, OpenCL SDK and WDK, and #1247 requires rebuilding the SDK bridge too).

I've pushed a combined branch with both fixes (#1246 + #1247) if you're able to build locally:
https://github.com/mansiverma897993/GenieX/tree/test/serve-openai-compat (build steps in notes/build.md)

@alanzhuly could CI produce a Windows ARM64 test artifact from a PR branch, or could these fixes go into the next pre-release build? That would let Novatoris verify both fixes end-to-end on Snapdragon hardware.

mengshengwu:
@mansiverma897993 Hi Mansi, thanks for your efforts. Apologies for the delayed reply.
The model manager has recently undergone refactoring, so this PR is now outdated. I will follow up on this task later.
To resolve the issue, the model manager needs to support the model:quant format, consistent with llama.cpp.

mengshengwu:
Superseded by #1261, which addresses this at the manifest layer (keys QAIRT ModelFile by the precision label directly) rather than adding an alias to the resolver — closing this in favor of that approach.

mansiverma897993:
> MAN$I VERMA (@mansiverma897993) Hi Mansi, thanks for your efforts. Apologies for the delayed reply. The model manager has recently undergone refactoring, so this PR is now outdated. I will follow up on this task later. To resolve the issue, the model manager needs to support the model:quant format, consistent with llama.cpp.

Np this is up to you ....and apart from these I have raised 2 more PR when you have chance take a look on it too .

PR #1248, opened 2026-07-28
fix(sdk): select clip backend from the resolved device, not ngl

### LINKED ISSUE #1178: Qwen3.5-0.8B GGUF crashes on IQ-9075 NPU with QGPUInstructionValidator assertion
## Describe the bug

Running `unsloth/Qwen3.5-0.8B-GGUF` with the GenieX NPU backend on a Dragonwing IQ-9075 EVK consistently crashes.

The same model runs successfully with the CPU backend.

When `--compute npu` is used, the process aborts inside the Qualcomm Adreno shader compiler with a `QGPUInstructionValidator` assertion.

The issue is reproducible on Qualcomm Linux Reference Distro 2.0.

---

## To Reproduce

1. Install GenieX on a Dragonwing IQ-9075 EVK.

2. Verify the GenieX version:

```bash
geniex version
```

Output:

```text
GenieX CLI Version:     v0.3.14
QAIRT Runtime Version:  v2.45.0.260326
LlamaCPP Runtime Hash:  be4a6a6
```

3. Download the model:

```bash
geniex pull unsloth/Qwen3.5-0.8B-GGUF
```

The selected chipset is:

```text
Dragonwing IQ-9075 EVK
```

The model downloads successfully and is automatically detected as:

```text
Detected model type: vlm
```

4. Verify that CPU inference succeeds:

```bash
geniex infer unsloth/Qwen3.5-0.8B-GGUF \
  --compute cpu \
  --ngl 0 \
  --think=false \
  --max-tokens 128 \
  --prompt "Please list five common inventory anomalies."
```

CPU inference completes successfully.

Observed performance:

```text
Approximately 29 tokens/s
Approximately 0.1 seconds to first token
```

A Traditional Chinese prompt also works correctly:

```bash
geniex infer unsloth/Qwen3.5-0.8B-GGUF \
  --compute cpu \
  --ngl 0 \
  --think=false \
  --max-tokens 128 \
  --prompt "請使用繁體中文列出五項常見庫存異常。"
```

5. Run the same model using the NPU backend:

```bash
geniex infer unsloth/Qwen3.5-0.8B-GGUF \
  --compute npu \
  --ngl 999 \
  --think=false \
  --max-tokens 128 \
  --prompt "Please list five common inventory anomalies."
```

6. The process crashes with the following error:

```text
NumConstRegsError: # of constant registers must be < 2.
    In inst# 1: %R6_c<def> = IMINrr %C22.x, %UR0_a, 1

Assertion failed: false && "back-end instruction validation failed",
file /usr/src/debug/qcom-adreno/1.855.4/shadercompiler/llvm/lib/Target/Oxili/QGPUInstructionValidator.cpp,
line 440

FATAL ERROR: false && "back-end instruction validation failed"
```

---

## Expected behavior

The model should run successfully using the IQ-9075 NPU backend and generate a text response.

If this model, VLM configuration, or one of its operators is not supported by the NPU backend, GenieX should return a clear unsupported-model or unsupported-operator error instead of terminating with a shader compiler assertion.

---

## Screenshots

Not applicable.

The full terminal output is included above.

---

## Environment

- Device: Dragonwing IQ-9075 EVK
- SoC: Qualcomm QCS9075
- Architecture: aarch64
- OS: Qualcomm Linux Reference Distro 2.0
- Kernel: `6.18.21-g7368c42b9cca-dirty`
- GenieX CLI: v0.3.14
- QAIRT Runtime: v2.45.0.260326
- LlamaCPP Runtime Hash: be4a6a6
- Adreno shader compiler version: 1.855.4
- Model: `unsloth/Qwen3.5-0.8B-GGUF`
- Precision: Q4_0
- Model type detected by GenieX: VLM

### OS information

```text
ID=qcom-distro
NAME="Qualcomm Linux Reference Distro"
VERSION="2.0"
VERSION_ID=2.0
PRETTY_NAME="Qualcomm Linux Reference Distro 2.0"
CPE_NAME="cpe:/o:openembedded:qcom-distro:2.0"
```

### Kernel information

```text
Linux iq-9075-evk 6.18.21-g7368c42b9cca-dirty #1 SMP PREEMPT Tue May 5 09:29:10 UTC 2026 aarch64 GNU/Linux
```

---

## Additional context

- The model runs correctly using `--compute cpu`.
- The failure occurs only when hardware acceleration is enabled.
- The crash occurs inside the Adreno shader compiler even though `--compute npu` is explicitly selected.
- Qwen3.5-0.8B is automatically detected as a VLM.
- The test is text-only and does not include an image input.
- The same issue was previously reproduced on Qualcomm Linux 1.6 with Adreno version 1.808.11.
- Reflashing the board to Qualcomm Linux Reference Distro 2.0 upgraded the Adreno stack to 1.855.4, but the same failure remains.

Could you please confirm:

1. Is `unsloth/Qwen3.5-0.8B-GGUF` Q4_0 officially supported by the GenieX NPU backend on IQ-9075 / QCS9075?
2. Is text-only NPU inference supported when this model is detected as a VLM?
3. Why does `--compute npu` fail inside the Adreno shader compiler?
4. Is this `NumConstRegsError` a known issue?
5. Is there a recommended GenieX version, model export, GGUF file, or BSP configuration for this model?

sanoop-sadheerthan:
## Summary

Any VLM GGUF crashes at model load on QCS9075 with **no flags at all**, because
the default compute unit is `npu`:

```
sh-5.2# geniex infer unsloth/Qwen3.5-0.8B-GGUF
NumConstRegsError: # of constant registers must be < 2.
    In inst# 1:    %R6_c<def> = IMINrr %C22.x, %UR0_a, 1
Assertion failed: false && "back-end instruction validation failed", file /usr/src/debug/qcom-adreno/1.855.2/shadercompiler/llvm/lib/Target/Oxili/QGPUInstructionValidator.cpp, line 440
FATAL ERROR: false && "back-end instruction validation failed"
```

Two things about this crash are worth noting up front, because both are clues:

1. It happens with **no `--compute` flag**. `geniex_resolve_device` defaults an
   empty alias to `npu`, so the plain command above is already the NPU path.
   Every user hits this by default — it is not an opt-in edge case.
2. It happens on a **text-only prompt**, before any image is supplied. So the
   failure is at *load*, not during image processing.

The assertion comes from the Adreno **shader compiler** — meaning something
asked the OpenCL/GPU backend to build kernels, even though the user asked for
the NPU.

## Why this happens

`unsloth/Qwen3.5-0.8B-GGUF` is a vision-language model: it ships
`mmproj-BF16/F16/F32.gguf`, so `is_mmproj_filename` in
`sdk/model-manager/crates/core/src/manifest_builder.rs` classifies it as
`ModelType::Vlm`. `mmproj_path` is therefore populated and the load goes through
`LlamaVlm::create`, not `LlamaLlm::create`.

Inside `LlamaVlm::create`, the language model is correctly pinned to the NPU —
`resolve_devices("HTP0")` sets `llama_model_params::devices = {HTP0}`. But the
**vision projector's** backend was chosen from a completely different signal:

```cpp
mparams.use_gpu = config.n_gpu_layers != 0;   // sdk/plugins/llama_cpp/src/vlm.cpp:85
```

This reads only the layer count and never looks at which device was requested.
The resolved alias table in `sdk/src/device.cpp` makes that fatal, because three
of the four aliases carry the same default `ngl = -1`:

| alias | `device_id` | `ngl` | `classify_device` | `use_gpu` before | correct |
|---|---|---|---|---|---|
| `cpu` | NULL | 0 (forced) | `CPU` | false | false |
| `gpu` | `GPUOpenCL` | -1 | `GPU` | true | true |
| **`npu`** | **`HTP0`** | **-1** | **`NPU`** | **true** ← bug | **false** |
| **`hybrid`** | NULL | -1 | `NPU` | **true** ← bug | **false** |

So on `npu`, clip was told `use_gpu = true`. clip then picks its own backend,
with no knowledge of the `{HTP0}` selection made 40 lines earlier
(`tools/mtmd/clip.cpp`):

```cpp
if (ctx_params.use_gpu) {
    ...
    backend = ggml_backend_init_by_type(GGML_BACKEND_DEVICE_TYPE_GPU, nullptr);
```

That instantiates the Adreno OpenCL backend and compiles its kernels, which
trips `QGPUInstructionValidator`. And because `mtmd_init_from_file` runs inside
`create()`, it fires at load time — which is exactly why a text-only prompt
crashes too.

**Where the regression came from.** `git blame` points at f54afca4 *"fix(sdk):
enable clip GPU backend when n_gpu_layers is -1"*, which changed `> 0` to
`!= 0`. That correctly fixed `--compute gpu` — whose default `ngl = -1` had made
the old test false — but the new test also captures `npu` and `hybrid`, which
carry the same `-1`.

## How this is fixed

`LlamaVlm::create` already computes the resolved compute target on its first
line; the decision just wasn't using it. One line:

```diff
--- a/sdk/plugins/llama_cpp/src/vlm.cpp
+++ b/sdk/plugins/llama_cpp/src/vlm.cpp
@@ -82,7 +82,7 @@ int32_t LlamaVlm::create(const geniex_VlmCreateInput* input) {
     // Initialize vision context if mmproj_path provided
     if (input->mmproj_path) {
         mtmd_context_params mparams = mtmd_context_params_default();
-        mparams.use_gpu             = config.n_gpu_layers != 0;
+        mparams.use_gpu             = device == Device::GPU;
         mparams.print_timings       = false;
         mparams.n_threads           = 4;
```

`const Device device = classify_device(input->device_id, config.n_gpu_layers)`
is already in scope from line 42. Now only `--compute gpu` selects OpenCL, so
`npu` never reaches the Adreno compiler.

Scope notes:

- `use_gpu` appears **exactly once** in the tree and is not plumbed through
  `geniex_ModelConfig`, so no binding can set it and there is no second
  instance of this bug to fix.
- `n_gpu_layers` still reaches `llama_model_params` untouched — this moves only
  the vision projector, not language-model layer offload.
- `--compute gpu` is unchanged (`classify_device` → `Device::GPU` → `true`), and
  `--compute cpu` is unchanged (`ngl == 0` → `Device::CPU` → `false`, as before).
  `--compute gpu --ngl 0` is also unchanged, because `classify_device` tests
  `n_gpu_layers == 0` first and returns `CPU`, matching the old expression.
- **One deliberate behaviour change beyond the four aliases.** Raw device ids
  bypass the alias table: `bindings/python/geniex/auto.py:91` passes
  `device_map="<plugin>:<id>"` straight through to `device_id`, with `ngl` left
  at the caller's `-1`. For those inputs `use_gpu` changes:

  | `device_map` | `classify_device` | `use_gpu` before → after |
  |---|---|---|
  | `llama_cpp:GPUOpenCL` | `GPU` | true → true (unchanged) |
  | `llama_cpp:HTP2` | `NPU` | true → **false** |
  | `llama_cpp:NOSUCHDEV` | `CPU` (unknown → CPU fallback) | true → **false** |

  Both changes are intended, not incidental: an unrecognised device should not
  silently instantiate OpenCL, and a non-pinned HTP index should not pull in the
  Adreno backend for exactly the same reason `HTP0` should not. On this board
  (#1250) the old behaviour meant a typo'd device id crashed the load.
- `sdk/include/geniex.h` is untouched, so this is not an FFI change.
- The `qairt` plugin has its own `vlm.cpp` sharing no code with this file, and
  its `create` rejects non-zero `ngl` outright — that path is unaffected.

## Test plan

Built and verified on real hardware: an **Innodisk EXMP-Q911** COM-HPC Mini
Module — Qualcomm **QCS9075 / Dragonwing IQ-9075** —
<https://www.innodisk.com/en/products/computing/qualcomm-solution/exmp-q911>

```
$ adb shell uname -a
Linux exmp-q911 6.6.119-qli-1.8-ver.1.1-06201-g380a343250d3 aarch64 GNU/Linux
$ adb shell cat /etc/os-release
NAME="Qualcomm Linux"   VERSION="1.8-ver.1.1"   ID=qcom-wayland
```

SDK cross-built with `cmake --preset arm64-linux-snapdragon-debug` in
`qualcomm/geniex-toolchain-linux:v0.1.0`; CLI with
`bazel build //cli --config=linux_arm64 --//sdk:sdk_type=local`. The unmodified
v0.3.17 install was baselined **first**, then the patched build was side-loaded
alongside it so before/after is a like-for-like comparison on one board.

Models: `unsloth/Qwen3.5-2B-GGUF:Q4_K_M` (VLM) and `Qwen/Qwen3-0.6B-GGUF:Q8_0`
(text-only). Image: `tests/assets/quality_dog.jpg`.

| model | compute | before | after |
|---|---|---|---|
| VLM, text-only prompt | `npu --ngl 999` | **crash (assertion above)** | **OK — 7.7 tok/s** |
| VLM + image | `npu --ngl 999` | **crash** | **OK — 7.9 tok/s, correct caption** |
| text-only model | `npu --ngl 999` | OK 35.2 tok/s | OK 37.9 tok/s |
| VLM, text-only prompt | `cpu --ngl 0` | OK 21.1 tok/s | OK 20.4 tok/s |
| text-only model | `cpu --ngl 0` | OK 42.0 tok/s | OK 41.3 tok/s |
| text-only model | `gpu --ngl 999` | crash | crash (pre-existing , #1250) |
| VLM + image | `gpu --ngl 999` | crash | crash (pre-existing, #1250) |

- [x] **Mechanism confirmed from the runtime log, not inferred.** With the fix,
      `GENIEX_LOG=debug` shows the projector landing on CPU:
      ```
      clip_ctx: CLIP using CPU backend
      mmproj loaded: vision=true, audio=false
      ```
      OpenCL is now only *enumerated*, never instantiated — no kernel build, no
      assertion:
      ```
      ggml_opencl: selected platform: 'QUALCOMM Snapdragon(TM)'
      ggml_opencl: device: 'QUALCOMM Adreno(TM) 663 (OpenCL 3.0 Adreno(TM) 663)'
      ```
      The same log also confirms the VLM classification and the NPU pinning:
      `mmproj_path: .../mmproj-F32.gguf`, `device_id: HTP0`, `n_gpu_layers: 999`,
      `resolve_devices] Using 1 device(s): HTP0`.
- [x] `pytest tests -m api` on device: `test_resolve_device_map` 6/6,
      `test_init_deinit` 3/3, `test_runtime_list` 2/2, `test_version` 2/2 —
      results byte-identical against the unmodified libs, so no regression.
- [x] `bazel test //cli/... //bindings/go/... --//sdk:sdk_type=local`: 9 pure-Go
      targets pass.
- [x] Compiles clean on two toolchains: `arm64-linux-snapdragon-debug` (cross)
      and the native `default` preset.

### Limitations, stated plainly

- **`--compute hybrid` is *not* fixed by this change.** It also crashes on a
  text-only model: `hybrid` sends an empty `device_id`, so
  `llama_model_params::devices` is never set and llama.cpp selects the broken
  OpenCL backend for the *language model*. Separate defect in the LLM path,
  deliberately left out of this PR.
- **`--compute gpu` and `--compute hybrid` stay broken on this board; this PR
  does not fix them — see #1250.** Both die in the same Adreno assertion, and
  both reproduce on a *text-only* model with no clip involved at all, which is
  how I separated them from this bug. Root cause is an upstream llama.cpp OpenCL
  kernel (`kernel_cpy_f32_f32_pack`, `ggml-opencl/kernels/cpy.cl:212`), not
  GenieX. Three consequences here:
  - `npu` no longer routes the vision projector into that path — that is #1178.
  - `hybrid` still crashes, by a different route: it leaves `device_id` empty, so
    `llama_model_params::devices` is never set and llama.cpp selects OpenCL for
    the *language model*. Out of scope for this PR.
  - The "`--compute gpu` still runs the projector on GPU" no-regression case
    could not be observed, because `--compute gpu` cannot complete a load on this
    board at all. It holds by inspection — `classify_device` returns
    `Device::GPU`, so `use_gpu` is `true` both before and after — but it was not
    demonstrated.

Closes #1178


mansiverma897993:
Thanks @sanoop-sadheerthan for verifying on device and raising this the change matches the root cause I traced in #1178 (clip's backend was chosen from `n_gpu_layers` while the LM was pinned to HTP0, so `npu`/`hybrid` with the default `ngl = -1` dragged in the Adreno OpenCL backend at load). The hardware before/after table and the `CLIP using CPU backend` log confirmation look solid. A few things I think are still needed:

1. **Guard this line with a comment.** This is now the third value this expression has had (`> 0` → `!= 0` in f54afca4 → `device == Device::GPU` here), and the previous regression happened precisely because the constraint wasn't written down. Suggest something like:

   ```cpp
   // Clip must follow the *resolved device*, not n_gpu_layers: npu/hybrid
   // carry the default ngl = -1, and using ngl here drags the Adreno OpenCL
   // backend into an NPU run (#1178). Only an explicit GPU device may use it.
   mparams.use_gpu = device == Device::GPU;
   ```

2. **One unstated behavior change in the scope notes:** for an *unknown* `device_id` with `ngl != 0`, the old code gave `use_gpu = true`, the new code gives `false` (via `classify_device`'s CPU fallback). I think that's an improvement an unrecognized device shouldn't silently instantiate OpenCL but it should be stated in the PR description so it's a deliberate decision, not an accident.

3. **A regression test.** Nothing automated currently pins the alias → clip-backend mapping, which is why f54afca4 could regress `npu` unnoticed. Even a small unit test asserting `classify_device` output for the four alias rows in your table (`cpu`/`gpu`/`npu`-`HTP0`/`hybrid`-empty, each with their default ngl) would prevent a fourth rewrite of this line from breaking it again.

Also, since `--compute hybrid` is explicitly left broken here (LM path picks OpenCL when `devices` is never set) is there an issue tracking that? Happy to take it as a follow-up; it falls straight out of the same trace. 
@sanoop-sadheerthan I can't merge it directly But I reviewd it ...
@alanzhuly @mengshengwu can you plz also review it I already done with review and posting that changes still required here .

sanoop-sadheerthan:
Thanks @mansiverma897993 — all four done.

**1. Guard comment.** In `1fc7f8c3`, your wording verbatim. Only oddity: it came out as a 6-line diff, because clang-format's `AlignConsecutiveAssignments` doesn't align across comments — so once a comment splits them, the three `mparams.*` lines re-align.

**2. Unknown `device_id`.** Updated the PR description to call it out as intentional. It's reachable too, not just theoretical: `auto.py:91` passes a raw `device_map="<plugin>:<id>"` straight through to `device_id`. I added `llama_cpp:HTP2` while I was there, since a non-pinned HTP index had the same problem.

**3. Regression test.** Added `plugins/llama_cpp/tests/test_params_device.cpp` — your four alias rows, the two raw-id rows from above, and the `ngl = 0` precedence case. Kept it in the style of `test_model_manager.c` rather than pulling in a framework.

I didn't want to just trust a green run, so I mutation-tested it: flipping `classify_device`'s unknown-device fallback to `GPU` makes it fail on exactly that row. It also checks that the old `n_gpu_layers != 0` expression still *disagrees* with the resolved-device answer, and fails if that ever stops being true — so going back to `ngl` can't look harmless.

<details>
<summary>Test output</summary>

```
ok   cpu                    device_id=(null)       ngl=  0 -> use_gpu=false
ok   gpu                    device_id=GPUOpenCL    ngl= -1 -> use_gpu=true
ok   npu                    device_id=HTP0         ngl= -1 -> use_gpu=false
ok   hybrid                 device_id=             ngl= -1 -> use_gpu=false
ok   llama_cpp:HTP2         device_id=HTP2         ngl= -1 -> use_gpu=false
ok   llama_cpp:NOSUCHDEV    device_id=NOSUCHDEV    ngl= -1 -> use_gpu=false
ok   gpu --ngl 0            device_id=GPUOpenCL    ngl=  0 -> use_gpu=false
note n_gpu_layers != 0 is wrong for npu                    (would give use_gpu=true)
note n_gpu_layers != 0 is wrong for hybrid                 (would give use_gpu=true)
note n_gpu_layers != 0 is wrong for llama_cpp:HTP2         (would give use_gpu=true)
note n_gpu_layers != 0 is wrong for llama_cpp:NOSUCHDEV    (would give use_gpu=true)
7/7 rows passed
```

With the mutation applied:

```
FAIL llama_cpp:NOSUCHDEV    device_id=NOSUCHDEV    ngl= -1 -> use_gpu=true, want false
6/7 rows passed
```

</details>

Digging into why `f54afca` got through turned up something I wasn't expecting. There's already a `llama_cpp_vlm` leg in `_qdc-pytest.yml` covering the `npu` cell — but it's commented out in `pr-check.yml` (`TODO(#976)`), so it never runs on PRs. And even switched on it wouldn't have helped: it runs on QDC X Elite, where Adreno OpenCL works fine, so clip loads onto OpenCL happily and the test goes green with the bug sitting right there. This only blows up where the shader compiler is broken.

Long way of saying I don't think any end-to-end test on the hardware CI currently has could have caught this — which is the real argument for pinning the mapping. Fair warning though: nothing invokes `ctest` in this repo today, so the test runs locally and for native-preset builds, but not in CI. I'd rather wire that into `_test.yml` (native arm64) as a separate `ci` PR than grow this one, if that works for you.

**4. Hybrid.** #1250. It covers `gpu` and `hybrid` together since they're the same root cause — `kernel_cpy_f32_f32_pack` in `ggml-opencl/kernels/cpy.cl:212` does `min(ne00, get_local_size(0))`, mixing a uniform register with a constant register, and Adreno's validator won't have it. I compiled all 166 kernels on the device one at a time; `cpy.cl` is the only casualty. Happy to pick the fix up as a follow-up.

@zhiyuan8  @mengshengwu @alanzhuly

mengshengwu (approved):
@sanoop-sadheerthan @mansiverma897993 
Thanks for your work, this did fix the issue.

mengshengwu:
Verified on QCS9075 / Dragonwing IQ-9075 EVK (Qualcomm Linux 1.9-ver.1.0, aarch64). Model: `unsloth/Qwen2.5-VL-3B-Instruct-GGUF` (Q4_K_M + mmproj-F16). Same command, same board, only `libgeniex_plugin.so` swapped between runs.

`geniex infer unsloth/Qwen2.5-VL-3B-Instruct-GGUF --compute npu --max-tokens 40 -p "..."`

| plugin | result |
|---|---|
| `mparams.use_gpu = config.n_gpu_layers != 0` (pre-fix) | crash — `NumConstRegsError` / `QGPUInstructionValidator.cpp:440` |
| `mparams.use_gpu = device == Device::GPU` (fix, 5b590dd) | 10.6 tok/s, 0.7 s first token |
| same fix after 0245bdc4 (test + comment trim) | 8.8 tok/s, 0.8 s first token |

Crash text on this board matches the PR description verbatim (driver `qcom-adreno/1.855.3` vs 1.855.2 in the PR — driver point-release only).

Issue #1266, opened 2026-07-31
qairt: intermittent fixed ~1.5s latency per graphExecute on X Elite (decode 0.64 tok/s vs 17)


NathanL15:
**Setup:** GenieX CLI v0.3.18 (QAIRT 2.45), Windows 11 ARM64, Snapdragon X Elite (X1E78100), 16 GB RAM, NPU driver 30.0.220.3000. Model: `ai-hub-models/Qwen3-4B-Instruct-2507` (same behavior with a LOCALFS import of the bundle).

**Symptom:** the qairt path flips between two states across process launches:

- healthy: ~17–19 tok/s decode, ~1,230–1,340 tok/s prefill (matches perf.yaml)
- degraded: ~0.64 tok/s decode, ~80 tok/s prefill

**This is not a low NPU clock — it is a fixed ~1.5 s penalty per `graphExecute` call.** qairt prefills in 128-token chunks and decodes 1 token per pass, so per-invocation cost is `prompt_time / ceil(n_prompt/128)` and `decode_time / n_gen`:

| state | ms per prefill pass (128 tokens) | ms per decode pass (1 token) |
|---|---|---|
| healthy | 95–104 | 51–58 |
| degraded | 1580–1664 | 1555–1624 |

Healthy, a prefill pass does 128× the work of a decode pass for ~2× the time — compute-bound, as expected. Degraded, the two cost **the same**: invocation cost becomes independent of how much work the pass does. Fitting `degraded = healthy + C` gives C ≈ 1498 ms from prefill and C ≈ 1514 ms from decode, and it predicts the measured `-p 512` prefill to within 0.1% (4 × (104 + 1500) = 6.40 s vs 6.407 s measured).

A DCVS/voltage-corner problem would scale both passes by the same *factor* and preserve the 2:1 ratio. The ratio instead collapses to 1:1, which only an additive per-invocation constant does — so NPU compute appears to be running at full speed even when degraded, and each invocation pays ~1.5 s of fixed latency. That points at FastRPC/DSP wake, sleep-latency / rpc-polling settings, or a timeout-and-retry, rather than at clock corners. (~1.5 s is a suspiciously timeout-shaped number.) 17 runs, data and analysis script available on request.

[...]

PR #1267, opened 2026-07-31
ci(release): make release SDK builds hermetic (skip cross-run caches)

NathanL15:
Adds a `hermetic` input to `_build-sdk.yml` that skips the three cross-run cache restores (ccache, cargo target, CMake build tree), and sets it for the `build-sdk` call in `release.yml`. Defaults to `false`, so PR/branch builds and every other caller keep their caches; only release artifacts pay the clean-build cost (one ~30–60 min build per release, inside the job's 120-minute timeout).

Rationale: shipped artifacts shouldn't depend on cross-run mutable state. The cargo-target cache comment in this file already records one stale-artifact incident in this pipeline; a hermetic release build removes that entire class for the binaries users install, whatever the mechanism of any individual incident.

Scope note, to be upfront: I verified that `actions/cache` extracts with archived mtimes (no `-m`), so ninja's staleness detection should normally be sound — this is defense-in-depth for releases, not a fix for a demonstrated bug, and it is independent of #1266 (which it came out of investigating). If you consider the current caches sufficient, closing this is a reasonable call.

### DIFF
--- a/.github/workflows/_build-sdk.yml
+++ b/.github/workflows/_build-sdk.yml
@@ -18,6 +18,17 @@ on:
         required: false
         type: string
         default: ""
+      hermetic:
+        description: >-
+          Build from clean state: skip every cross-run cache restore (ccache,
+          cargo target, CMake build tree). Release builds set this so shipped
+          binaries can never be assembled from stale cached outputs — the same
+          uniform-mtime hazard documented on the cargo cache below applies to
+          the ninja build tree, whose cache key hashes only configure inputs,
+          not the C/C++ sources.
+        required: false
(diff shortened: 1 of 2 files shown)


mengshengwu:
Thanks for the suggestion. We’ll decline this change.

These caches are build‑time only and cannot taint final shipped artifacts. They are reliable and critical for CI performance; disabling them for releases would add 30‑60 min per build and hurt developer productivity.

This is defense‑in‑depth for hypothetical risk, and the performance cost outweighs the benefit. We can revisit if real cache‑caused release issues surface.

PR #1352, opened 2026-08-16
fix(cli): run update installer silently

### LINKED ISSUE #1344: CLI: geniex update silently without pop window
When Geniex update is executed, uninstall and installation pop‑up windows will appear, causing the update to hang in environments without a user interface. In addition, the update approach requiring user interaction provides a poor user experience.

mansiverma897993:
## Summary

Add a `geniex update --silent` flag that launches the Windows installer with `/VERYSILENT /SUPPRESSMSGBOXES /NORESTART` so upgrades work in headless/CI environments without pop-up windows. Default `geniex update` behavior is unchanged.

- `geniex update` → launches the installer normally (existing behavior).
- `geniex update --silent` → launches the installer silently.
- Setup uses `SuppressibleMsgBox` for the "existing version detected" confirm and always execs the previous uninstaller with `/VERYSILENT /NORESTART /SUPPRESSMSGBOXES`. Bare `MsgBox` under `/VERYSILENT` still renders as a hidden dialog and would deadlock setup — the extra flags also silence the old uninstaller's own dialogs, which `/VERYSILENT` alone does not.
- Removed the `update_test.go` unit test (real update flow is exercised at release time).
- The `exec.Command` target is a SHA256-verified file downloaded from the official release S3; noted with a `nosemgrep` marker so the preflight scanner treats it as reviewed.

Closes #1344

## Test plan

Local Windows ARM64 host, upgrading real installed CLI each time:

- [x] **Silent upgrade from a pre-PR install.** `v0.3.19` → `0.4.0-silentfix` via `Start-Process -Wait <installer> /VERYSILENT /SUPPRESSMSGBOXES /NORESTART`. Returns in ~7s; install log shows `Existing version detected. Uninstall now?` auto-answered under `/VERYSILENT`, previous uninstaller invoked and cleared, new uninstall key registered.
- [x] **Silent-loop upgrade (upgraded-from-silent → silent again).** `0.4.0-silentfix` → `0.4.0-silentfix2` via the same flags. ~6s, same log pattern. This is the real production upgrade path.
- [x] **`/SILENT` mode (progress bar visible, no confirms).** `0.4.0-silentfix2` → `0.4.0-silentfix3`. Confirms `SuppressibleMsgBox` auto-answers under `/SILENT` as well as `/VERYSILENT`.
- [x] **Full CLI path end-to-end.** Pointed the CLI at a local range-aware HTTP release channel serving a `0.4.0-silentfix4` manifest + installer, ran `geniex update --silent` from the freshly-built CLI: manifest fetched, chunked parallel download SHA256-verified, `exec.Command(...).Start()` fired the installer with the three silent flags, CLI exited 0, `geniex version` on PATH became `0.4.0-silentfix4` ~10s later. Local URL patch reverted after the test.
- [x] **Pre-fix hang reproduced on the same host** to confirm the failure mode: unpatched v0.4.0 installer under `/VERYSILENT` left `geniex-cli-setup-*.exe` + `.tmp` idle (~0.2s CPU across 60s), zero writes to `%LOCALAPPDATA%\GenieX CLI\`, hidden `#32770 'Setup'` window pending on the `InitializeSetup` `MsgBox`.
- [x] `bazelisk test //cli/...` locally: 15 passed / 6 skipped (posix-only, expected).
- [x] Interactive `geniex update` (no `--silent`) on a host with a prior version — expected to show the setup wizard while the previous uninstaller runs invisibly (matches pre-PR interactive UX). Not re-verified on this host to avoid interrupting the desktop; the `SuppressibleMsgBox` default-value path is only used in silent mode, interactive keeps stock `MsgBox` semantics.


cli/cmd/geniex/update_test.go
mengshengwu, line 1: We rely real update test instead of text matching, So I think this tests code is unnecessary

cli/cmd/geniex/update.go
```diff
@@ -138,13 +138,17 @@ func runUpdate(_ *cobra.Command, _ []string) error {
...
 
-	if err := exec.Command(dst).Start(); err != nil {
+	if err := updateInstallerCmd(dst).Start(); err != nil {
 		return err
 	}
 	fmt.Println("update package is ready to install")
 	return nil
 }
 
+func updateInstallerCmd(dst string) *exec.Cmd {
```
mengshengwu: Only once call, so just inline it.

mengshengwu (changes requested):

cli/cmd/geniex/update.go
mengshengwu, line 1: It would be better add a `geniex update --silent ` flag let update silently and not change default behavior

mengshengwu (changes requested):

mengshengwu (approved):

PR #1353, opened 2026-08-17
feat(sdk): resolve AIHM version from latest.txt instead of pinning

Hahahaooovvv:
## Summary
- Model Manager now resolves the AIHM release version via `releases/latest.txt` instead of a pinned constant, so `geniex pull`/`list` can pick up new AIHM releases without a code change.
- Falls back to the pinned default on any failure (network, missing pointer, malformed content) — never blocks a pull.
- Reuses the same 24h disk cache as manifest.json/platform.json, so a version bump still cascades into invalidating those.
- `GENIEX_AIHUBVERSION` still overrides outright, skipping the network entirely.

## Why draft
AIHM hasn't actually published `latest.txt` yet — qcom-ai-hub/ai-hub-models-internal#4284 merged, but no release has run since (still 403 on S3 as of today). Verification so far is limited to mocked scenarios (unit/integration tests + a manual smoke run); haven't run it against the real pointer yet. Will validate live and mark ready once AIHM's next release publishes it.

Refs: qcom-ai-hub/geniex#1434

### DIFF
--- a/sdk/model-manager/crates/core/src/source/ai_hub/mod.rs
+++ b/sdk/model-manager/crates/core/src/source/ai_hub/mod.rs
@@ -22,6 +22,7 @@ use std::time::{Duration, SystemTime};
 use async_trait::async_trait;
 use url::Url;
 
+use crate::config::StoreConfig;
 use crate::error::{parse_manifest, Error, Result};
 use crate::manifest::{ModelFileInfo, ModelManifest, ModelType};
 use crate::transport::{HttpTransport, ReqwestTransport};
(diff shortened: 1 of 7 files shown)


sdk/model-manager/crates/core/src/source/ai_hub/mod.rs
```diff
@@ -38,9 +39,12 @@ use super::{basename, BytesSource, FileSpec, ModelSource, Plan};
...
 const MANIFEST_FILENAME: &str = "manifest.json";
 const PLATFORM_FILENAME: &str = "platform.json";
+const LATEST_VERSION_FILENAME: &str = "latest.txt";
 const MAX_INDEX_BYTES: u64 = 8 * 1024 * 1024;
 
-/// TTL for the on-disk index cache (matches Go CLI's 24h).
+/// TTL for the on-disk index cache (matches Go CLI's 24h). Also used for
+/// the `latest.txt` pointer cache — AIHM releases roughly monthly, so a
+/// separate, shorter TTL for the pointer wasn't worth the extra constant.
 const CACHE_TTL: Duration = Duration::from_secs(24 * 60 * 60);
```
vinovo, line 46: Is there currently a way to force disabling the cache? If not, we need to reduce this to 1h. We can't let users to wait too long to get the latest AIHM release artifacts.

RemiliaForever (approved):

PR #1372, opened 2026-08-21
feat(sdk): fetch AIHM index from releases/latest/, pinned as fallback

Hahahaooovvv:
## Summary

- Replace the `releases/latest.txt` pointer with direct fetches from `releases/latest/{manifest,platform}.json` — the new AIHM contract (qcom-ai-hub/geniex#1434 rev-2, discussed with @gustavla / @paulzhu / @khoi).
- Fall back to `releases/<cfg.version>/...` when the `latest/` fetch fails, so GenieX keeps working before AIHM cuts the first release under the new layout.
- `GENIEX_AIHUBVERSION` still short-circuits to the pinned version outright.
- Cache TTL: `manifest.json` / `platform.json` = 1h (mtime-only, no more version-stamp cross-check); `info.json` stays at 24h.

## Test plan

- [x] `cargo test -p model-manager-core --lib --tests` — 168 lib + 3 pull + 3 release_index + 5 executor + 5 proxy_smoke + 3 local_aihub_pull, all green
- [x] Live S3 (`--ignored live_manifest_resolves_phi_model`) — `latest/` currently 403 → transparent fallback to `v0.60.0`
- [x] `geniex model list --all` against real S3: WARN on 403 + falls back + lists 19 models
- [x] `geniex model list --all` against local range-aware mock serving `latest/manifest.json`: no WARN, latest/ hit
- [x] `GENIEX_AIHUBVERSION=v0.60.0 geniex model list --all`: `latest/` skipped, no WARN

Refs: qcom-ai-hub/geniex#1434


vinovo (approved):

sdk/model-manager/crates/core/src/source/ai_hub/mod.rs
```diff
@@ -39,18 +39,21 @@ use super::{basename, BytesSource, FileSpec, ModelSource, Plan};
...
 
-/// 24h TTL for manifest / platform / info JSON caches (matches Go CLI).
-const CACHE_TTL: Duration = Duration::from_secs(24 * 60 * 60);
+/// TTL for `latest/` index JSONs — short so new releases propagate.
+const INDEX_TTL: Duration = Duration::from_secs(60 * 60);
 
-/// 1h TTL for the `latest.txt` version pointer.
-const LATEST_VERSION_TTL: Duration = Duration::from_secs(60 * 60);
+/// TTL for per-model info.json (immutable per release).
+const INFO_TTL: Duration = Duration::from_secs(24 * 60 * 60);
```
vinovo, line 50: @Hahahaooovvv @RemiliaForever Why should info.json ever be invalidated and updated?

My understanding is that this is tied to a model bundle, and I don't see reason to update it unless the bundle is re-pulled or deleted?

PR #1389, opened 2026-08-25
feat(qairt): bundled QAIRT runtime by default, --qnn-lib as an optional override

zhycheng614:
## Summary

Adds `--qnn-lib` / `GENIEX_QNN_LIB`: point the SDK at a QAIRT SDK root or a flat folder of QNN libraries to run against a QAIRT version other than the bundled one. This is what lets us hot-swap the QAIRT version on a single GenieX release instead of cutting a new build.

- **Optional override, not a requirement.** A QAIRT runtime stays bundled, so default behavior is unchanged.
- **One build spans QAIRT versions.** The plugin reaches QNN only through the versioned C interface, which negotiates at load time — so there is no ABI variant to select and no per-version build matrix.
- **Accepts either layout**: a full QAIRT SDK root (host libs under `lib/<triple>`, DSP skels under `lib/hexagon-v*/unsigned`) or a flat folder holding the libs directly.
- **Packaging back to a single `lib/qairt`** + `lib/qairt/htp-files`, with the `htp-files` install restored.
- **Fails fast** on a path that is missing or holds no QNN libs, naming the recovery.
- Requires the plugin side: qualcomm/geniex-qairt-plugin#40.

Verified end to end on Snapdragon X Elite (v73) through `bazel run //cli`, for both an LLM and a VLM, across QAIRT 2.45 / 2.48 / 2.49 — including switching back and forth and returning to the bundled runtime.

<details>
<summary>What changed</summary>

The submodule bump is what forces the rest: at the old pin `QAIRT_QNN_VERSION` was a real cache option with per-version header sets; it no longer exists. Left alone, the per-version CI loop would build identical binaries and `selectQairtVariant()` would skip every `qairt-*` directory whenever `GENIEX_QNN_LIB` was unset — the plugin would never load at all.

- **Bridge** (`qnn_runtime_utils.h`) — unset leaves all path fields `nullopt` and the plugin self-resolves from the bundled `htp-files/`. Set, we resolve here, because the plugin only understands the flat shape and translating an SDK root is our job. Also corrects `#if defined(WIN32)` to `_WIN32`.
- **`registry.cpp`** — drops `selectQairtVariant` / `detectQnnApiMinor` / `qairtVariantForMinor` and the `qairt-*` skip (85 lines, plus `<fstream>`/`<regex>`).
- **CLI** — `--qnn-lib` documented as optional; sets `GENIEX_QNN_LIB` for the process before SDK init, so the LLM and VLM paths and any in-process binding pick it up uniformly.
- **`notes/run.md`** — the ABI-unsafe warning is replaced (the C API negotiates), and the verification tip now names the line the plugin actually logs when an override is active.
- **CI / `release.yml` / `_sdk_fetch.py`** — reverted to the single-build form.

</details>

<details>
<summary>Test matrix</summary>

Qwen3-4B-Instruct-2507 and Qwen2.5-VL-7B:

| Case | Result |
|---|---|
| Install layout | single `lib/qairt` + `htp-files` (38 files), no `qairt-<ver>` dirs |
| No flag / no env | self-locates the bundled runtime |
| Flat folder 2.45 / 2.48 | loads and generates |
| QAIRT SDK root 2.48 | host libs resolved to `lib/aarch64-windows-msvc`, skels found |
| `GENIEX_QNN_LIB` env alone | works |
| Flag + env both set | flag wins |
| 2.48 → 2.45 → 2.48 → bundled | all pass |
| VLM text + image, bundled / 2.45 / 2.48 / 2.49 | override reaches the vision encoder too |
| Nonexistent path / dir with no libs | fails fast in ~3 s with the documented message |

</details>

<details>
<summary>Two findings, neither caused by this branch</summary>

- On the multi-CL 4B bundle, QAIRT 2.49 hits the known marginal PD ceiling (`err 1002` on context 2). It reproduces outside the CLI with no SDK involved, and the same bundle passed earlier on the same machine, so it is not a 2.49 regression.
- After any failed load the process takes an access violation (`0xC0000005`) in teardown and then hangs instead of exiting. Plugin-side, worth a separate issue.

</details>

Issue: qcom-ai-hub/geniex#1203


cli/cmd/geniex/infer.go
```diff
@@ -141,6 +143,14 @@ func infer() *cobra.Command {
 			return err
 		}
 
+		// Exported rather than passed down so every path in this process -- LLM, VLM, any
+		// binding -- picks the override up the same way.
+		if qnnLib != "" {
+			if err := os.Setenv("GENIEX_QNN_LIB", qnnLib); err != nil {
+				return fmt.Errorf("failed to set GENIEX_QNN_LIB: %w", err)
+			}
+		}
```
vinovo, line 152: Several questions 
1.. `geniex-qairt` also takes the lib path as a runtime config, right? Why don't we pass it all the way down to `cgo`?
2.. How do you plan to have it supported for `geniex serve`?
3. Did you consider which UX is better - using `geniex config` versus passing this as a param to `geniex infer` on each run?

notes/run.md
```diff
@@ -109,6 +109,65 @@ Warning: qairt plugin only supports NPU inference; ignoring device='cpu' and run
...
+a mismatch could segfault (ai-hub-models-internal#3964). That hazard is gone with the C API,
+which is why this is a supported override rather than a testing-only aid.
+
+`QnnHtpNetRunExtensions` is no longer loaded at all — the plugin applies
+`htp_backend_ext_config.json` itself through the public C API — so a runtime folder does not
+need to carry it.
+
+With no override in play the plugin logs `HTP runtime path: … (auto-resolved from bundled
+htp-files/)` instead; that line is absent whenever an override is active, because the pinned
+paths short-circuit the plugin's own resolver.
```
vinovo: Recommend removing. I don't think this concerns most GenieX users

sdk/plugins/qairt/include/qnn_runtime_utils.h
```diff
@@ -29,61 +54,128 @@ inline std::optional<std::string> find_optional_file(const std::filesystem::path
...
+    GENIEX_LOG_DEBUG("Setting ADSP_LIBRARY_PATH to {}", adsp_path);
+#if defined(_WIN32)
+    _putenv_s("ADSP_LIBRARY_PATH", adsp_path.c_str());
+    SetDllDirectoryA(host_dir.string().c_str());
+#else
+    setenv("ADSP_LIBRARY_PATH", adsp_path.c_str(), 1);
 #endif
+    runtime_cfg.backend_path    = (host_dir / kQnnBackendLib).string();
+    runtime_cfg.system_lib_path = (host_dir / kQnnSystemLib).string();
+    runtime_cfg.extensions_path = (host_dir / kQnnExtensionsLib).string();
```
vinovo: I think only one of the two should happen:
1. `geniex-qairt` takes a single folder path that expects certain layout, and in the sdk layer, it directly passes the env or the folder path to `geniex-qairt` without any handling or validation. That means the path resolution logic should be in `geniex-qairt` and we need a new field in the struct `RuntimeConfig`
2. The sdk layer does the path resolution and environment variable resolution, like you have in this PR. Then `geniex-qairt` should not need to handle any environment / path resolution.

In https://github.com/qualcomm/geniex-qairt-plugin/pull/40 you have similar ADSP and environment resolution logic. We should choose from one of the option and remove the duplication.

zhycheng614: Agreed, and I'd pick **option 1** — the plugin owns resolution. The duplication is worse than it looks: the two layers currently *disagree*.

| | SDK (`qnn_runtime_utils.h`) | plugin (`core/include/runtime.h`) |
|---|---|---|
| reads `GENIEX_QNN_LIB` | `:167` | `selectHtpDir` rung 2 |
| sets `ADSP_LIBRARY_PATH` | always | only if `hasHexagonSkels(dir)` |
| accepts a QAIRT SDK root | yes, via `lib/<triple>` | **no — errors out on it** |

So the plugin tells you *"This looks like a QAIRT SDK root. Point `GENIEX_QNN_LIB` at the host library directory itself … not the SDK root"* for the same variable this layer documents as accepting an SDK root. Which contract applies depends on which layer resolves first, so a plugin-direct user and a GenieX user get opposite answers.

Option 2 costs more than it looks: the plugin ships standalone, documents `GENIEX_QNN_LIB` in its own README, and unit-tests the four-rung precedence, none of which involves this SDK. Option 1 only asks the plugin to accept a layout it already detects well enough to reject.

This is a cross-repo change, so I've left it out of this PR and will do it as two follow-ups. Fine with the split?

<details><summary>Why the SDK-side ADSP code isn't currently redundant, and the follow-up shape</summary>

`resolveHtpPaths` early-returns when all three path fields are set (`runtime.h:277`), and this layer pins all three — so on the override path the plugin never runs its resolver and never sets `ADSP_LIBRARY_PATH`. The SDK copy is compensating for a short-circuit it caused. Worth knowing before deleting either half.

**Follow-up (a), plugin:** widen the resolver to accept a QAIRT SDK root as well as a flat folder, so layout translation and ADSP live in one place for both.

**Follow-up (b), here:** plumb the path through the C API into `QnnRuntimeConfig::htp_dir` (the field already exists), reduce `make_qnn_runtime_config` to a forward, and keep the plugin's env rung as the fallback. That also answers your cgo question on the `infer.go` thread — and fixes the bindings, which cannot reach this knob today: `geniex_ModelConfig` has no field for it, so the env var is the only route, and Android has no `setenv` from Kotlin at all.

</details>

notes/run.md
zhycheng614: Removed in `5c2c365b`. Agreed — plugin internals, and it only existed to explain the duplication you flagged on the `qnn_runtime_utils.h` thread, so it goes away with that anyway.

cli/cmd/geniex/infer.go
zhycheng614: 1. **Yes, and we should.** `QnnRuntimeConfig::htp_dir` already exists in the plugin; the blocker is that `geniex_ModelConfig` has no field to carry it, so the env var is the only route down. That's the same conclusion as your `qnn_runtime_utils.h` thread, so I've folded it into follow-up (b) there rather than doing it here. It also fixes the bindings, which cannot reach this knob today — Android has no `setenv` from Kotlin, so an APK is stuck with the bundled runtime regardless.

2. **Fixed in `bc68bd21`.** `serve` keeps its own flag set rather than sharing `flagGroups`, so the flag was on `infer` and `run` but not the server; the env var always worked, since both end at the same plugin call. `GENIEX_QNN_LIB` is bound explicitly, because `AutomaticEnv` would derive `GENIEX_QNNLIB` from the viper key and quietly make it a second spelling.

3. **Per-run, not persisted** — but say the word and I'll add the key. Reasoning below.

<details><summary>On `geniex config` vs a per-run param</summary>

`store.ConfigKeys` holds exactly one key today, `chipset` — a stable property of the machine, cheap to be wrong about. A QAIRT runtime override is the opposite: a short-lived compat/debugging action whose whole risk is being forgotten. A stale pin is the one failure mode we know bites hard here — a mismatched runtime loads and runs at full speed while generating garbage, so there's no crash to tell you the pin is stale. Persisting it turns a per-command choice into a durable footgun, and `--qnn-lib` on the command line is self-documenting in a way `config.json` is not.

The counter-argument I can see is repeated benchmarking across versions, where retyping the path each run is friction. If that's the real use case I'd rather solve it with a shell alias than a persisted key.

</details>

sdk/plugins/qairt/include/qnn_runtime_utils.h
vinovo: If your plugin expects a certain folder layout and file names, you'd need to document it somewhere to make it transparent to users, and clearly state the failure mode

cli/cmd/geniex/infer.go
vinovo: Where does the impression of "QAIRT runtime override is for debugging" action? Compat / perf is the real use case here.

Retyping on each run is a big burden and forgetting means falling back to 2.45, as bad as using a stale version. I can see the downside of using config only being not able to easily switch version for a single model. That being said, perhaps we should support both geniex config which sets the default lib location and allow per-run override to take precedence.

zhycheng614: You're right on both counts, and the second one flips my argument. Implemented in `75c94d27`.

**"Debugging" was wrong** — compat and perf are the real uses. That framing is what led me to the wrong conclusion, so I've dropped it from the docs too.

**Not persisting has the same failure mode I was warning about.** Forgetting `--qnn-lib` silently drops back to the bundled 2.45; a stale stored pin silently keeps an old one. Both are "silently on the wrong runtime", so abstaining buys nothing — the answer is precedence, which is what you proposed:

```
--qnn-lib  >  GENIEX_QNN_LIB  >  geniex config set qnn-lib  >  bundled
```

A stored default never outranks what you typed for this run.

```bash
geniex config set qnn-lib /path/to/qairt/2.XX.0
geniex config get qnn-lib
```

<details><summary>Implementation and the visibility problem it leaves</summary>

- `store.ConfigKeyQnnLib` joins `ConfigKeys`, so `config get/set/list` and key validation pick it up with no per-key wiring.
- `resolveQnnLib(flag, env, config)` is pure and covered by 8 table tests, so the order is asserted rather than inferred; `applyQnnLib` does the store read and the SDK call. `infer` and `serve` share both.
- A config read failure is logged at debug and treated as unset, so a corrupt `config.json` cannot break a run that passed `--qnn-lib` explicitly.

**Worth knowing:** a stored default is still invisible at a glance — nothing prints "using your configured runtime" at the top of a run. What does exist is the INFO line naming the resolved directory, and `geniex config get qnn-lib`. If you want the stored case called out louder — say a one-line notice on stdout whenever the runtime came from config rather than the flag — that's a small follow-up and I think it's a good idea.

</details>

sdk/plugins/qairt/include/qnn_runtime_utils.h
zhycheng614: Fair — it was documented only in `notes/run.md`, which no user reads. `grep -rn "qnn-lib\|GENIEX_QNN_LIB" docs/` returned nothing before this. Fixed in `75c94d27`.

**[CLI reference](https://github.com/qualcomm/GenieX/blob/AIHW/workbench-build/docs/en/run/cli/reference.mdx)** now has a *Running against a different QAIRT runtime* section covering:

1. Both accepted layouts, with the file and folder names each one is recognised by.
2. The accepted version range — 2.36 and up, and why.
3. The precedence chain, including the new config default.
4. **The failure text, verbatim**, so it is searchable when someone hits it.
5. A warning that a wrong-but-accepted runtime produces bad output at full speed, i.e. that a successful run is not evidence the override resolved where you meant.

Python and Android API references carry the new calls, Android noting that the API is its only route. Chinese docs mirrored. `mint export` in `docs-check` will validate the MDX on this push.

<details><summary>The layout and failure mode, as documented</summary>

| Layout | Recognised by |
|---|---|
| QAIRT SDK root | host libs under `lib/<target-triple>/`, skels under `lib/hexagon-v*/` |
| Flat folder | `QnnHtp.dll` / `QnnSystem.dll` (or `libQnn*.so`) directly inside |

```
geniex: QnnHtp.dll not found in HTP runtime folder: D:/wrong/path
  (from QnnRuntimeConfig::htp_dir)
Expected either a flat directory holding QnnHtp.dll and its arch stubs (shaped like the
bundled htp-files/), or a QAIRT SDK root with lib/aarch64-windows-msvc/.
```

The C API doc comment for `geniex_set_qairt_runtime_path` now names both layouts and the version floor too, so an embedder reading only the header gets the contract.

Two things I did **not** do, in case you want them: the plugin's own README documents the layouts but not this failure text (that repo, separate PR), and `docs/en/resources/troubleshooting.mdx` has no entry for it — the reference section felt like the right single home, but troubleshooting is where someone with the error in hand would look.

</details>

vinovo (approved):

vinovo:
Addressed both points, @mengshengwu:

## 1. Dropped the C API and its downstream wiring

Reverted the process-global \geniex_set_qairt_runtime_path\ / \geniex_get_qairt_runtime_path\ C API (PR #1423) and everything built on it, back to the plain \GENIEX_QNN_LIB\-env-var mechanism this PR originally shipped with:

- \46cfdf5a\ -- reverts the C API in \geniex.h\/\ml.cpp\, the Go/Python/Android setter-getter bindings, the CLI's switch to calling the C API instead of exporting the env var, the \qnn_runtime_utils.h\ refactor that stopped resolving the env var itself, the \qnn-lib\ config default, and the plugin submodule bump those depended on.

Back to: SDK resolves \GENIEX_QNN_LIB\ / \--qnn-lib\ entirely in this repo's qairt plugin layer, no public C API surface added.

## 2. Renamed GENIEX_QNN_LIB -> GENIEX_QAIRT_LIB

- \6fb7d3d6\ -- renames it in this repo (CLI flag help text, env var name, error messages, \
otes/run.md\, \qnn_runtime_utils.h\)
- Plugin-side rename opened as qualcomm/geniex-qairt-plugin#53 (\kEnvVar\, doc comments, README, test)
- \4a74a43\ -- bumps the submodule pin to that plugin commit so both halves of the rename land together

Went with \GENIEX_QAIRT_LIB\ over \QAIRT_LIBRARY_PATH\: keeps the \GENIEX_\ prefix consistent with every other env var this project defines (\GENIEX_LOG\, \GENIEX_HFTOKEN\, \GENIEX_LIB_PATH\, etc.), while still naming the *runtime* being selected rather than just the QNN libraries inside it.

#53 will need to merge and the submodule pin to move once more (squash SHA), same pattern as the earlier #40/#51 bumps on this PR.

Davidqian123 (approved):

PR #1393, opened 2026-08-26
feat(qairt): switch to QAIRT_LIBRARY_PATH, restore default bundled behavior

mengshengwu:
Restores out-of-the-box QAIRT inference with one bundled QAIRT 2.45 runtime set, while allowing users to override the runtime location with `QAIRT_LIBRARY_PATH`.

- Renames `GENIEX_QNN_LIB` to `QAIRT_LIBRARY_PATH` and removes the `--qnn-lib` CLI flag
- Removes runtime selection among precompiled QAIRT plugin variants
- Updates the geniex-qairt submodule to the C API based single-build implementation from #41
- With `QAIRT_LIBRARY_PATH` unset, loads the bundled main-branch QAIRT 2.45 runtime
- With `QAIRT_LIBRARY_PATH` set, loads QNN host and HTP libraries from the user-provided QAIRT installation
- Re-installs the plugin build-tree `htp-files/` output into `lib/qairt-2.45/htp-files`

## Test plan
- [x] C API based SDK and CLI build successfully with the bundled-runtime configuration
- [x] Installation produces one bundled `qairt-2.45/htp-files` runtime set
- [x] Real `qualcomm/Qwen3-4B` inference with the bundled 2.45 runtime on HTP (23.5 tok/s, no crash)
- [x] Real `qualcomm/Qwen3-4B` inference with `QAIRT_LIBRARY_PATH` pointing to an external QAIRT 2.48 runtime on HTP (23.7 tok/s, no crash)


mengshengwu:
## Architecture note for reviewers

While testing this PR I found that `geniex-qairt-plugin`'s `AIHW/workbench-build` branch has since moved past the multi-variant approach this PR builds on (`23d3613 feat(qnn): multi-version QNN ABI, no vendored libs`, pinned via the submodule bump here). Its current tip (`8616c77 feat(qnn): reach QNN only through the C API, single build for every QAIRT`, `5d24253`) replaces the per-ABI-bucket build entirely:

- `CMakeLists.txt` no longer has a `QAIRT_QNN_VERSION` option -- it hardcodes one header set (`qnn-api/qnn-versions/2.45/include`), used only for type/struct definitions
- `QnnApi.cpp` loads the backend via `dlopen()` + `dlsym("QnnInterface_getProviders")`, i.e. QNN's own stable, versioned C API -- not a header-linked C++ vtable

That means a single build can, in principle, load *any* QAIRT runtime the user points `QAIRT_LIBRARY_PATH` at, with no precompiled per-version plugin variants (`qairt-2.45`/`2.47`/`2.48`) needed at all -- superseding the `selectQairtVariant()` dispatch this PR restores default behavior for.

Not changing course here since validating the C-API migration's stability/completeness is a separate, larger effort -- flagging so it's a known factor when deciding whether to keep maintaining the multi-variant dispatch long-term or rebase onto the C-API single-build design once it's ready.

mengshengwu:
## Design clarification

I updated this PR to the single-build layout: GenieX packages one default QAIRT runtime set, and `QAIRT_LIBRARY_PATH` overrides where the QNN host and HTP libraries are loaded from. The old `qairt-2.45` / `qairt-2.47` / `qairt-2.48` plugin dispatch has been removed.

Your finding in geniex-qairt-plugin#40 is important for the compatibility boundary: the current `IBackend` C++ interface is not stable across QAIRT releases because its vtable and method signatures change. Therefore this PR should not claim that the current plugin binary can safely load every QAIRT version. The single-binary cross-version guarantee belongs to the C-API migration in #40; this PR keeps the smaller packaging and runtime-path change while that work remains separate.

mengshengwu:
## Final design

This PR now follows the combined design: #41 supplies the C API based single-build plugin, and this PR packages only the main branch's QAIRT 2.45 runtime as the default. `QAIRT_LIBRARY_PATH` remains the override for a complete external QAIRT installation.

The previous per-version plugin dispatch and `qairt-2.47` / `qairt-2.48` packaging assumptions are removed. The C API migration is now consumed through the geniex-qairt submodule rather than reimplemented in GenieX.

mengshengwu:
## Test results (on device, Snapdragon X Elite / HTP)

Both scenarios verified end-to-end with the C API based single-build plugin (no crash, real generated output):

| Scenario | `QAIRT_LIBRARY_PATH` | Result |
|---|---|---|
| Default bundled runtime | unset | Loads `qairt-2.45/htp-files`; `qualcomm/Qwen3-4B` runs on HTP, 23.5 tok/s |
| External runtime override | `C:\...\real-qairt-248` (genuine QAIRT 2.48 host + HTP v73 libs) | Log confirms `Using QNN libraries from QAIRT_LIBRARY_PATH`; same plugin binary loads QAIRT 2.48 and runs on HTP, 23.7 tok/s |

No `0xC0000005` in either case. This confirms the C API migration in #41 removes the cross-version `IBackend` ABI issue previously hit when overriding to a newer QAIRT version.

PR #1401, opened 2026-08-27
fix(cli): harden the model-manager call surface

RemiliaForever:
## Summary

Six defects around the model manager's lifecycle and its `name:precision` keys,
found by an audit of every `geniex_model_*` call site in the CLI, server and Go
binding. They share a blast radius — a model reference that silently resolves to
the wrong thing, or an error the user never sees — so they land together.

**`ModelInit` failure was invisible, then fatal much later.** The root
`PersistentPreRun` logged with `slog.Error` and carried on. The default log level
is `none`, so an unwritable or corrupt data directory produced *no output at
all*; every later call then failed as `NOT_INITIALIZED`, an error that points
nowhere near the cause. It is `PersistentPreRunE` now and returns
`initialize model manager at <path>: <err>`.

**`ModelInit` had no counterpart.** A `PersistentPostRun` calls `ModelDeinit`,
placed next to the init it mirrors. Unchecked and unguarded on purpose: the SDK's
`geniex_model_deinit` is a no-op that always returns `GENIEX_SUCCESS`, which is
also why cobra skipping `PostRun` on error costs nothing today.

**`store.init()` reported fatal errors on bare stderr.** Two duplicated
`fmt.Fprintf(os.Stderr, "geniex: ...")` + `os.Exit(1)` pairs, unthemed, with no
hint at the fix. One `fatal()` helper now logs *and* prints through the theme,
and names `--data-dir` / `GENIEX_DATADIR`.

**`name + ":" + quant` was hand-assembled in 7 places**, each with its own
`if quant != ""`. New `geniex_sdk.JoinNamePrecision` — the exact inverse of the
existing `SplitNamePrecision`, minus its URL strip — replaces all of them.

**The server hid the `N/A` precision it already accepts.** `RetrieveModel`
matches `:N/A` on input via `EqualFold`, but both handlers stripped the suffix
from the id they returned, so a model with an undetermined precision was
advertised as `org/name` — which round-trips to a *different*, SDK-picked
reference. Both now return `org/name:N/A`.

`PrecisionNA`'s doc comment was wrong about what it means and is corrected here:
the AI Hub / QAIRT layout files every entrypoint under `N/A` and swaps in the
real label once known, so a surviving `N/A` means **undetermined**, not
**unquantized**. That mistake is what produced the two handler special cases.

**`geniex_model_resolve_alias` had an FFI entry but no Go binding** — Python has
`_maybe_resolve_alias`, Go could not reach the table. Exported as `ResolveAlias`.

Not in this PR, deliberately:

- `geniex list` still hides `N/A` in its PRECISIONS column
  (`downloadedPrecisions(model, !verbose)`), so the CLI table and `/v1/models`
  now disagree for an undetermined-precision model. That is the rest of audit 21.
- `store.init()` keeps its own `os.Exit(1)`, which fires *before* `ModelInit`, so
  the data-directory error story still has two paths.

## Test plan

- [x] `bazelisk build //cli --test_env=PATH --action_env=PATH`
- [x] `bazelisk coverage //cli/... --combined_report=lcov --test_env=PATH --action_env=PATH` — 15 passed / 6 skipped
- [x] `bazelisk test //bindings/go:geniex_sdk_go_test --test_env=PATH --action_env=PATH` — new `TestResolveAlias` + `TestJoinNamePrecision`
- [x] `gofmt -l cli/ bindings/go/` — clean
- [x] `:N/A` is a real reference, not a display string: with a hand-written manifest carrying `"ModelFile": {"N/A": {...}}`, `geniex infer qualcomm/NaTest:N/A` resolves the path and reaches `geniex_llm_create` (failing only on this host's missing qairt plugin), while `:Q4_0` fails earlier with `quantization 'Q4_0' not found`
- [x] Against a live `geniex serve`, `GET /v1/models`, `GET /v1/models/qualcomm/NaTest` and `GET /v1/models/qualcomm/NaTest:N%2FA` all return `"id": "qualcomm/NaTest:N/A"`
- [x] `geniex --data-dir /proc/nope model list` prints the themed `Error: create data directory ...` plus the `--data-dir` hint, exit 1

Closes qcom-ai-hub/geniex#1518

### DIFF
--- a/bindings/go/model_manager_test.go
+++ b/bindings/go/model_manager_test.go
@@ -1,0 +1,23 @@
+// Copyright 2024-2026 Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: BSD-3-Clause
+
+package geniex_sdk
+
+import "testing"
+
+// No ModelInit: the alias table is static, so no store is involved.
+func TestResolveAlias(t *testing.T) {
+	got, ok := ResolveAlias("qwen3")
+	if !ok {
+		t.Fatal("qwen3 is the SDK's alias fixture")
+	}
+	if want := "ggml-org/Qwen3-1.7B-GGUF:Q4_K_M"; got != want {
+		t.Errorf("got %q, want %q", got, want)
+	}
+
+	for _, alias := range []string{"", "no-such-alias-xyz-9999", "ggml-org/Qwen3-1.7B-GGUF"} {
+		if got, ok := ResolveAlias(alias); ok {
+			t.Errorf("ResolveAlias(%q): expected miss, got %q", alias, got)
+		}
+	}
+}
--- a/bindings/go/model_manager_test.go
+++ b/bindings/go/model_manager_test.go
@@ -21,3 +21,22 @@ func TestResolveAlias(t *testing.T) {
 		}
 	}
 }
+
+func TestJoinNamePrecision(t *testing.T) {
+	cases := []struct{ name, precision, want string }{
+		{"org/repo", "Q4_0", "org/repo:Q4_0"},
+		{"org/repo", "", "org/repo"},
+		// N/A is a manifest key like any other; hiding it is the caller's job.
+		{"org/repo", PrecisionNA, "org/repo:N/A"},
+	}
+	for _, c := range cases {
+		if got := JoinNamePrecision(c.name, c.precision); got != c.want {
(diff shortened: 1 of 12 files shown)


bindings/go/model_manager_test.go
mengshengwu, line 1: Do we real need this test file on go?

mengshengwu (approved):

PR #1418, opened 2026-08-31
feat(sdk): re-introduce ModelScope hub for geniex pull

### LINKED ISSUE #1085: ModelScope integration status — confirm whether to re-introduce it
### Body:

Hi maintainers 👋

I noticed that the ModelScope integration, which I previously added in #855 , no longer works on current main.

### Current status

The functional downloader was removed during the geniex refactor (last present in v0.2.73). Only an enum placeholder MODELSCOPE = 2 survives across the FFI bindings — there is no modelscope downloader under sdk/model-manager/crates/core/src/source/, and build_pull_intent() in pull.rs silently falls back to HuggingFace for ModelScope.

The CLI does not accept --model-hub modelscope at all:

```
$ geniex pull Qwen/Qwen3-0.6B-GGUF --model-hub modelscope
Error: unknown model hub: modelscope
```
(hf still works; modelscope/ms/s3/volces are all rejected.)

### Question

Is ModelScope still a hub you want supported in the geniex architecture, or was its removal intentional (e.g., superseded by another mechanism)? If the former, I'm happy to put up a PR re-integrating it against the new Rust model-manager architecture. Just wanted to confirm intent before starting the work.

Thanks!😊

mansiverma897993:
## Summary

Re-introduces ModelScope as a first-class hub in the Rust model-manager. The functional downloader was lost in the geniex refactor (#855); only the `MODELSCOPE = 2` enum placeholder survived, and `--model-hub modelscope` was rejected.

## Changes

- **`ModelScopeSource`** (`sdk/model-manager/crates/core/src/source/modelscope.rs`): lists files via `GET /api/v1/models/{repo}/repo/files` (`Recursive=true`, `Revision=master`), downloads each file via `GET /api/v1/models/{repo}/resolve/master/{path}` (redirects to the underlying Alibaba OSS object, so the existing range/resume machinery just works). Uses a repo-shipped `geniex.json` verbatim, else infers a manifest via `infer_manifest_from_names` (same as HuggingFace).
- ModelScope does **not** answer `HEAD` on its API endpoints, so the small JSON documents are fetched with plain GET through a dedicated client; weight sizes come from the listing so no HEAD is needed there either.
- **`PullIntent::ModelScope`** wired through `build_pull_intent` (FFI) and `build_source`.
- **CLI**: `--model-hub modelscope` (alias `ms`) now accepted.
- Honors the `MODELSCOPE_ENDPOINT` env var (same as the ModelScope SDK); strips pasted `https://modelscope.cn/models/...` URLs.

## Test report

- `cargo test -p model-manager-core --lib`: **185 passed** (includes 6 new ModelScope unit tests: inference, shipped manifest, nested paths / tree entries, 404, VLM via config.json, endpoint env override).
- Integration tests (`cargo test --tests`): all passed.
- `cargo clippy -p model-manager-core -p model-manager-ffi --all-targets`: clean (only pre-existing warnings in `ai_hub_release_index.rs`).
- `cargo fmt --check`: clean.
- **Live end-to-end vs real ModelScope** (`Qwen/Qwen3-0.6B-GGUF`): `end_to_end_pull_via_modelscope` and `end_to_end_pull_qwen3_q8_0` both pass full 639 MB Q8_0 download + manifest publish + `get_paths` resolution.
- Go: `gofmt` clean; CLI + bindings compile (full Go test run requires the Bazel SDK build, covered by CI).

Resolves #1085.

### DIFF
--- a/cli/cmd/geniex/model_test.go
+++ b/cli/cmd/geniex/model_test.go
@@ -122,6 +122,38 @@ func TestPrintListCSV(t *testing.T) {
 	}
 }
 
+func TestResolveHubModelScope(t *testing.T) {
+	modelHub = "modelscope"
+	defer func() { modelHub = "" }()
+	h, err := resolveHub()
+	if err != nil {
+		t.Fatalf("resolveHub: %v", err)
+	}
+	if h != geniex_sdk.HubModelScope {
+		t.Errorf("resolveHub = %v, want %v", h, geniex_sdk.HubModelScope)
+	}
+}
+
+func TestResolveHubModelScopeAlias(t *testing.T) {
+	modelHub = "ms"
+	defer func() { modelHub = "" }()
+	h, err := resolveHub()
+	if err != nil {
+		t.Fatalf("resolveHub: %v", err)
+	}
+	if h != geniex_sdk.HubModelScope {
+		t.Errorf("resolveHub = %v, want %v", h, geniex_sdk.HubModelScope)
+	}
+}
+
+func TestResolveHubUnknownRejected(t *testing.T) {
+	modelHub = "modelscope-not-a-hub"
+	defer func() { modelHub = "" }()
+	if _, err := resolveHub(); err == nil {
+		t.Error("resolveHub accepted an unknown hub, want error")
+	}
+}
+
 func TestSkipDownloaded(t *testing.T) {
 	candidates := []geniex_sdk.PrecisionCandidate{
 		{Precision: "Q4_0", Size: 100},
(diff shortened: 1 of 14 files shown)


cli/cmd/geniex/model_test.go
```diff
@@ -122,6 +122,38 @@ func TestPrintListCSV(t *testing.T) {
...
+}
+
+func TestResolveHubUnknownRejected(t *testing.T) {
+	modelHub = "modelscope-not-a-hub"
+	defer func() { modelHub = "" }()
+	if _, err := resolveHub(); err == nil {
+		t.Error("resolveHub accepted an unknown hub, want error")
+	}
+}
+
```
mengshengwu: These cases are belong to modle-manager instead of go, so it would be better rm these tests here

mengshengwu (changes requested):
LGTM. Do you test it with `geniex pull` command?

mengshengwu (approved):

PR #1420, opened 2026-09-01
fix(sdk): pin mtmd vision encoder to HTP for npu/hybrid

mengshengwu:
## Summary
- Pin the mtmd vision encoder (`clip`) to the resolved HTP device for the `npu` / `hybrid` device aliases, instead of relying on `use_gpu` + ggml's type-based backend auto-pick. ggml-hexagon registers as `GGML_BACKEND_DEVICE_TYPE_GPU`, so when OpenCL is also enabled, the auto-pick can land on OpenCL/Adreno instead of HTP, leaving clip unaccelerated for the default `npu` path.
- Fall back to CPU if HTP initialization for the vision encoder fails, so vision support degrades gracefully instead of being lost entirely.
- Bump the llama.cpp submodule to `b10731`. The previously pinned commit shares one eagerly-created HTP session between the LLM context and clip's own scheduler; on real HTP hardware this corrupts the session's async queue state across multiple VLM turns once clip actually runs on HTP (`dspqueue_read` failures). The bump picks up ggml-hexagon's "create sessions on demand" rework (upstream #27785), which this scenario needs.
- Refresh `llama-hexagon-release-sessions.patch` for the new per-device `ggml_backend_hexagon_device_context` wrapper (session is now a lazily-created `unique_ptr` instead of the raw context pointer), and adapt the two call sites the version jump broke: `mtmd_helper_bitmap_init_from_file` gained an `mtmd_helper_init_opt` parameter, and `common_chat_tools_parse_oaicompat` now takes the new opaque `common_json` type instead of `nlohmann::ordered_json`.
- Add `llama-opencl-adreno-xmem-gemm-off.patch`: the bump also picks up upstream #27640, which defaults ggml-opencl's Adreno xmem F16xF32 GEMM path on. On this Adreno generation it corrupts output after several `gpu`/`hybrid` model loads through the same OpenCL context — confirmed as a genuine bump regression by running the full `test_llama_cpp.py` suite against unpatched `main` (44 passed, 0 failed) vs. the bumped commit before this patch (35 passed, 9 failed, all `gpu`/`hybrid` quality/logits-parity checks). Defaulting it back off via upstream's own `GGML_OPENCL_ADRENO_XMEM_GEMM=0` override resolves it.

## Test plan
- [x] `--device npu`, SmolVLM-256M + mmproj + image: clip loads on `HTP0` (was `CPU`); TTFT drops from ~4.0s to ~0.6s on this host
- [x] `--device hybrid`: clip also loads on `HTP0`
- [x] `--device gpu`: unchanged — clip still loads on `OpenCL` (no regression)
- [x] Full SDK build with `GENIEX_PLUGIN_QAIRT=ON` alongside `llama_cpp` succeeds and the HTP pinning above still holds with both plugins loaded
- [x] `test_vlm_multi_turn[gemma-4-E2B-it-GGUF-npu]` — the scenario that crashed pre-bump (`dspqueue_read` on real HTP hardware) — passes locally (Windows ARM64) and on QDC android (SM8850), which now runs the full `test_llama_cpp.py` suite clean
- [x] Full `test_llama_cpp.py` suite (44 tests, cpu/gpu/npu/hybrid, MTP excluded) passes 100% locally with both patches applied, matching a clean `main` baseline run byte-for-byte in pass count

### DIFF
--- a/sdk/plugins/llama_cpp/src/vlm.cpp
+++ b/sdk/plugins/llama_cpp/src/vlm.cpp
@@ -90,14 +90,23 @@ int32_t LlamaVlm::create(const geniex_VlmCreateInput* input) {
 
     // Initialize vision context if mmproj_path provided
     if (input->mmproj_path) {
+        ggml_backend_dev_t htp_device = device == Device::NPU ? ggml_backend_dev_by_name("HTP0") : nullptr;
+
         mtmd_context_params mparams = mtmd_context_params_default();
-        mparams.use_gpu             = device == Device::GPU;
+        mparams.use_gpu             = device == Device::GPU || htp_device != nullptr;
+        mparams.device              = htp_device;
         mparams.print_timings       = false;
         mparams.n_threads           = 4;
         // Zack TODO: elegant fix this error:  no member named 'verbosity' in 'mtmd_context_params'
         // mparams.verbosity           = GGML_LOG_LEVEL_ERROR;
 
         this->ctx_vision = mtmd_init_from_file(input->mmproj_path, this->model, mparams);
+        if (!this->ctx_vision && htp_device) {
+            GENIEX_LOG_WARN("mtmd failed to initialize the vision encoder on HTP; falling back to CPU");
+            mparams.use_gpu  = false;
+            mparams.device   = nullptr;
+            this->ctx_vision = mtmd_init_from_file(input->mmproj_path, this->model, mparams);
+        }
         // Continue even if vision context fails
         if (this->ctx_vision) {
             this->supports_vision = mtmd_support_vision(this->ctx_vision);
--- a/sdk/plugins/llama_cpp/src/vlm.cpp
+++ b/sdk/plugins/llama_cpp/src/vlm.cpp
@@ -5,7 +5,6 @@
 
 #include <algorithm>
 #include <cstring>
-#include <nlohmann/json.hpp>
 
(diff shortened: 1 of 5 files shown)


sdk/plugins/llama_cpp/src/vlm.cpp
```diff
@@ -90,14 +90,23 @@ int32_t LlamaVlm::create(const geniex_VlmCreateInput* input) {
 
     // Initialize vision context if mmproj_path provided
     if (input->mmproj_path) {
+        ggml_backend_dev_t htp_device = device == Device::NPU ? ggml_backend_dev_by_name("HTP0") : nullptr;
+
         mtmd_context_params mparams = mtmd_context_params_default();
-        mparams.use_gpu             = device == Device::GPU;
+        mparams.use_gpu             = device == Device::GPU || htp_device != nullptr;
+        mparams.device              = htp_device;
```
RemiliaForever, line 96: should we just pass `selection->data()` as llm part?

mengshengwu: That's a good idea, but what users want is good MTMD performance. Now that the upstream bugs have been fixed, MTMD can directly use the same device (including GPU/NPU). Is this parameter not very necessary for now? Otherwise, we will also need to modify the SDK API and bindings. Let's add it later when there is a clear demand.

PR #1462, opened 2026-09-16
fix: parse MiniCPM5 XML tool calls

### LINKED ISSUE #1452: Tool calls not working with MiniCPM5 models: GenieX only parses OpenAI JSON format, not XML-style function calls
**Describe the bug**
When running a MiniCPM5 model via GenieX and connecting it to an AI agent (e.g., pi agent), tool/function calls completely fail. GenieX currently only implements parsing for the standard OpenAI JSON tool call format, but MiniCPM5 outputs tool calls in an XML-style format (`<function name="..."><param ...>`). GenieX does not perform the reverse conversion from this XML format to a structured tool call response, causing the agent to be unable to invoke any tools.
This is a known gap in GenieX — not a bug in the model itself.
 
**To Reproduce**
1. Start a GenieX server with a MiniCPM5 GGUF model: `geniex serve "openbmb/MiniCPM5-2B-GGUF:Q4_K_M" --compute gpu`
2. Connect the GenieX endpoint to an OpenAI-compatible AI agent that supports tool use (e.g., pi agent with a custom provider).
3. Send a prompt that requires the agent to invoke a tool (e.g., a file read, web search, or any registered tool).
4. See error — the tool call is never executed; the agent reports that the model's response could not be parsed as a valid tool call.
 
**Expected behavior**
GenieX should detect that the model has responded with an XML-style function call (MiniCPM5 format) and transparently convert it into the standard OpenAI JSON tool call structure before returning the response to the client. This would allow MiniCPM5 models served via GenieX to work seamlessly with any OpenAI-compatible agent or framework that relies on tool use.
 
**Screenshots**
N/A
 
**Desktop (please complete the following information):**
- OS: Windows (ARM)
- Browser N/A
- GenieX CLI Version: v0.6.1
- QAIRT Runtime Version: 2.45
- LlamaCPP Runtime Hash: 0eadefe
 
**Smartphone (please complete the following information):**
- Device: N/A
- OS: N/A
- Browser N/A
- Version N/A
 
**Additional context**
N/A

mansiverma897993:
## Summary

MiniCPM5 emits XML function calls whose `<function`, `</function>`, `<param`, and `</param>` delimiters are control tokens. The runtime previously discarded these delimiters before the HTTP tool-call scanner saw the output, so scanner-only tests passed while real requests returned text and `finish_reason: "stop"`.

- Parse the official MiniCPM5 XML grammar into OpenAI-compatible tool calls, including XML entities, CDATA, multiple calls, and streamed output.
- Preserve the four XML delimiter tokens during llama.cpp generation, for both streamed callbacks and complete responses. Detection uses the vocabulary token text and works with a neutral model filename.
- Add blocking/streaming response tests and a standalone real-model HTTP regression script, including tool-result follow-up turns.

Fixes #1452.

## Testing

Tested the complete HTTP flow with the official `openbmb/MiniCPM5-1B-GGUF` Q4_K_M model on an x86_64 CPU under Ubuntu/WSL. Built the PR sources and the pinned llama.cpp commit `0eadefebd3f8f92a86d634a0e5b8fffc9dc792c0`; CPU-only development build, QAIRT disabled. The model was imported as `local/regression:Q4_K_M` with filename `model.gguf`.

The original and fixed backends received identical blocking and streaming request bodies. Generation was greedy, without mocked output or a forced grammar.

| Case | Original backend | Fixed backend |
|---|---|---|
| Blocking tool call | Missing XML delimiters; no tool calls; `stop` | `get_weather({"city":"Beijing"})`; `tool_calls` |
| Streaming tool call | Missing XML delimiters; no tool calls; `stop` | `get_weather({"city":"Beijing"})`; `tool_calls` |
| Blocking tool-result follow-up | No call ID available | Supplied call ID accepted; answer includes 23°C; `stop` |
| Streaming tool-result follow-up | No call ID available | Answer includes supplied 23°C result; `stop`; SSE `[DONE]` |
| Ordinary chat without tools | `Hello` | `Hello` |

A native probe using the actual GGUF vocabulary confirmed that token IDs 18–21 decode to zero bytes with `special=false` and to the four XML delimiters with `special=true`.

Also passed:

```text
go test -p 2 -cover ./server/... -count=1
clang-format 18.1.8 check
ruff 0.6.9 check and format check
git diff --check
```

To reproduce with a built CLI, pull the model and start the server, then run the HTTP test in another terminal:

```sh
geniex pull openbmb/MiniCPM5-1B-GGUF:Q4_K_M
geniex serve --compute cpu --ngl 0
python tests/http/minicpm5.py --model openbmb/MiniCPM5-1B-GGUF:Q4_K_M --output minicpm5-http-results
```

Model SHA-256: `81b64d05a23b17b34c475f42b3e72fbde62d4b92cc34541f7a8031d0752deafa`.


mansiverma897993:
@RemiliaForever, fixes MiniCPM5 XML tool-call parsing. Could you please run a quick Windows ARM MiniCPM5 end-to-end tool-call check? Thanks!

RemiliaForever:
Have you tested it locally? This feature is not hardware-dependent, so local testing should be enough.

mansiverma897993:
> Have you tested it locally? This feature is not hardware-dependent, so local testing should be enough.

Sorry my bad , Ik this is fully software ...give me some time to retest it and will leave with proper command to test !!

mansiverma897993:
Retested locally and updated the PR with the exact commands and results. The complete tool-call scanner suite passed (0.489s), and all focused MiniCPM5 parse/streaming cases passed (0.565s). I also simplified the implementation to the official MiniCPM5 XML grammar only. The latest CI rerun is awaiting maintainer approval after the force-push; please approve it when convenient.  Sorry for previous confusion again my bad that silly mistake ...

RemiliaForever:
Hi, have you tested the complete HTTP request flow? When I tested with `openbmb/MiniCPM5-1B-GGUF`, I found that detokenize does not render the special tokens, which prevents the tool call from matching.

mansiverma897993:
> Hi, have you tested the complete HTTP request flow? When I tested with `openbmb/MiniCPM5-1B-GGUF`, I found that detokenize does not render the special tokens, which prevents the tool call from matching.

Ohh yes , thanks for catching this. I reproduced the full HTTP failure: MiniCPM5 XML delimiter tokens were suppressed during detokenization, so the parser received incomplete calls.  
I fixed it and verified real-model HTTP requests: blocking and streaming tool calls, tool-result follow-ups, and ordinary chat all pass locally. now check at once !!

